PR #24, opened 2019-05-02
Config form validation

loneil:
Add form validation on the UI components in the Config Form section.

# Description

Front end form field validation from subtasks in:
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-64

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation (non-breaking change with enhancements to documentation)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

## Further comments

Using built-in Vuetify validation. See examples at
https://vuetifyjs.com/en/components/text-fields
https://vuetifyjs.com/en/components/text-fields


bcgov-nr-csst:
Deployed to https://getok-pr-24-k8vopl-dev.pathfinder.gov.bc.ca

frontend/src/components/ConfigForm.vue
```diff
@@ -84,12 +137,36 @@ export default {
...
       commonServices: [
         { text: "Common Messaging Service", value: "cmsg" },
         { text: "Document Management Service", value: "dms" },
         { text: "Document Generation Service", value: "dgen", disabled: true }
       ],
-      userAppCfg: this.$store.state.userAppCfg
+      userAppCfg: this.$store.state.userAppCfg,
+      applicationAcronymRules: [
+        v => !!v || "Acroynm is required",
+        v => v.length <= 30 || "Acroynm must be 10 characters or less",
```
parc-jason: Length comparison should be 10 here.
v => v.length <= 10 || "Acroynm must be 10 characters or less",

parc-jason:
line 151 for application acronym has a length check of 30, and a message for 10 characters.  Should update that to match.

frontend/src/components/ConfigForm.vue
matthewhall78: There might be an app that has a acronym longer than 10 characters in existence right now I’m not sure that 10 is our max length but it might be long enough to cover off all existing app acronyms. I’ll go into IRS and do a quick check

matthewhall78: There might be an app that has a acronym longer than 10 characters in existence right now I’m not sure that 10 is our max length but it might be long enough to cover off all existing app acronyms. I’ll go into IRS and do a quick check

loneil: @parc-jason right, thanks for catching that discrepancy.
@matthewhall78 yeah I'd actually arbitrarily selected 10, need to find the real max length as you mentioned.

matthewhall78: currently this is the longest app name:
PRIVATE PROPERTY DB

matthewhall78: also IRS acronym list has some that contain lowercase, spaces, numbers and these non-alpha numerics:
 ().-_
for example
  ADAM (WILDFIRE) 
  SMARTTOOL 2.0

i think apps that have included there api version in the acronym made a mistake in doing so, since they will need to create a new IRS entry when a new version of the api is released.  so I don’t what to allow service client names or app acronyn names that end in V1 or 2.0.  but I dont see an issue with numbers being used like in app “A4CA”.

so we may need to add a backlog item to look at these edge cases. for now the stricter rule we have is what i want to go with.

loneil: Yeah restricting newly created ones to a standard is best in my opinion.

frontend/src/components/ConfigForm.vue
```diff
@@ -6,35 +6,71 @@
...
-      ></v-select>
+      <v-form v-model="step1Valid">
+        <v-layout row wrap>
+          <v-flex xs12 md7>
+            <v-text-field
+              label="Application Acronym"
+              required
+              :value="userAppCfg.applicationAcronym"
+              v-on:keyup.stop="updateAppCfgField('applicationAcronym', $event.target.value)"
+              :counter="30"
```
jujaga: Wondering if we should be making the counter a variable instead of a magic number across the codebase

frontend/src/components/ConfigForm.vue
```diff
@@ -43,67 +79,120 @@
...
+        >Submit</v-btn>
+        <v-btn flat @click="appConfigStep = 1">Back</v-btn>
+      </v-form>
     </v-stepper-content>
   </v-stepper>
 </template>
 
 <script>
 import { mapGetters } from "vuex";
+import {isValidJson} from "../utils/utils.js"
```
jujaga: spaces around isValidJson
Also if there's a way to use the @ relative path instead of directory traversal via .. it would make it cleaner

frontend/src/components/ConfigForm.vue
```diff
@@ -43,67 +79,120 @@
...
       commonServices: [
         { text: "Common Messaging Service", value: "cmsg" },
         { text: "Document Management Service", value: "dms" },
         { text: "Document Generation Service", value: "dgen", disabled: true }
       ],
-      userAppCfg: this.$store.state.userAppCfg
+      userAppCfg: this.$store.state.userAppCfg,
+      applicationAcronymRules: [
+        v => !!v || "Acroynm is required",
+        v => v.length <= 30 || "Acroynm must be 30 characters or less",
```
jujaga: Wondering if we should be making the counter a variable instead of a magic number across the codebase

frontend/src/components/ConfigForm.vue
```diff
@@ -43,67 +79,120 @@
...
+        v =>
+          /^(?:[A-Z]{2,}[_]?)+[A-Z]{2,}$/g.test(v) ||
+          "Incorrect format. Hover the ? for details."
+      ],
+      applicationNameRules: [
+        v => !!v || "Name is required",
+        v => v.length <= 30 || "Name must be 30 characters or <b>less</b>"
```
jujaga: Wondering if we should be making the counter a variable instead of a magic number across the codebase

frontend/src/utils/utils.js
```diff
@@ -0,0 +1,8 @@
+export function isValidJson(json) {
+    try {
+        JSON.parse(json);
+    } catch (e) {
+        return false;
+    }
+    return true;
+}
```
jujaga: This file isn't indented correctly
Add a newline at the end of file

frontend/src/components/ConfigForm.vue
loneil: Ah yes, I'd done '@/utils/utils.js' in the unit test file but forgot to adjust here.

PR #76, opened 2019-07-24
Add Patroni Postgresql DB support

jujaga:
# Description

This PR adds a Postgres database to the API backend. As this is a large infrastructure and design change, many aspects of the application and pipeline flow have been changed to compensate.
- **Database**
  - Create and leverage Patroni - a high availability implementation of Postgres (leveraged from bcgov image)
- **Pipeline**
  - Update Jenkins pipeline to handle patroni standup and secret generation
  - Use ephemeral patroni for PRs, and persistent for main deployments
  - Parallelize patroni standup wherever possible
  - Consolidate OC template labels
  - Add mid lifecycle hook to backend deployment to manage DB migrations
- **Security**
  - Remove snyk integration (no longer bcgov direction)
  - Update lodash (Advisory https://npmjs.com/advisories/1065)
- **Backend**
  - Add graceful failure support to backend (intercepts SIGTERM and SIGINT)
  - Force node backend to run in production mode
  - Log users into user table when they access system for the first time
- **Frontend**
  - Add gzip compression support to Caddy

This is required in order to fulfill the requirement of auditing tangible actions done on the application.

[[SHOWCASE-100]](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-100)

## Types of changes

Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

## Further comments

The following is the ER diagram of the database we will be starting with.
erdiagram 8-1-2019


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-76

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-76

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-76

backend/src/controllers/users.js
```diff
@@ -0,0 +1,12 @@
+const User = require('../models').User;
+
+module.exports = {
+  async findOrCreate(keycloakId, displayName) {
+    return await User.findOrCreate({
+      where: {
+        keycloakId: keycloakId,
+        displayName: displayName
```
parc-jason: Should displayName be in the where clause?  
keyCloakId is unique, but display name isn't.  I think you're going to run into issues if someone changes their display Name outside our db.  Move displayName into 'defaults: {displayName: displayName}

backend/src/migrations/20190729000000-initialize-database.js
```diff
@@ -0,0 +1,228 @@
...
+          },
+          acronymId: {
+            allowNull: false,
+            references: {
+              model: 'acronym',
+              key: 'acronymId'
+            },
+            type: Sequelize.UUID
+          },
+          appCfg: {
```
parc-jason: Should use a more easily understood name for this field...  If all the node code is using appCfg to call the WEBADE Rest API, then the class property name is ok, but the database column name should probably be changed to application_config or something clear.  Even app_config is better than 2 abbreviations

backend/src/models/user.js
```diff
@@ -0,0 +1,40 @@
...
+    keycloakId: {
+      allowNull: false,
+      comment: 'Associated Keycloak user id',
+      type: DataTypes.UUID,
+      unique: true,
+      validate: {
+        isUUID: 4
+      }
+    },
+    displayName: {
```
parc-jason: Should we be saving which system has defined the displayName?  BCEID, IDIR, GitHub?
One thing we should include in the model is which Keycloak instance this id is from...  Perhaps we need another model/table to track that...

openshift/backend.bc.yaml
```diff
@@ -2,19 +2,16 @@
 apiVersion: v1
 kind: Template
 labels:
+  build: "${REPO_NAME}-backend"
```
parc-jason: Just a heads up - there are name length limitations in OpenShift.  24 characters I believe.  Using repo name is not a big deal for nr-get-token, but other projects (ex. nr-messaging-service-showcase) takes up a lot of the available characters. So, probably best not to used repo name in any names or labels... (for next project)

backend/src/controllers/users.js
jujaga: Good call - updated in upcoming commit

backend/src/migrations/20190729000000-initialize-database.js
jujaga: I've gone ahead and changed it to appConfig - there's a lot existing code calling it as appConfig.

backend/src/models/user.js
jujaga: - I've just added a username field which will be Keycloak's representation of what the user and system is from (ie steve@idir or bob@github).
- I'm not sure why we'd need to include which instance of Keycloak an ID is coming from - there is only ever going to be one Keycloak that the application will authenticate to because it is the source of truth.
- There was some discussion about whether or not we should just consolidate the userId and keycloakId together to be one sole primary key. We could go either way and there are merits for different naming and structure. May need more discussion on that front.

openshift/backend.bc.yaml
jujaga: 👍

PR #84, opened 2019-08-21
Feature/get app config

loneil:
# Description
A backend endpoint implementation to allow fetching WebADE configurations via the getok api.
Restricts read access to acronyms that a user is associated with, or a power user can be in the WEBADE_CFG_READ_ALL group.

GET on /api/v1/appConfig/{webadeEnv}/{acronym}

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/store/modules/auth.js
+++ b/frontend/src/store/modules/auth.js
@@ -20,8 +20,9 @@ export default {
         const payload = JSON.parse(atob(token.split('.')[1]));
         const roles = payload.realm_access.roles;
 
+        // TODO: this will require re-conceptualizing as acronyms will come from the DB, not the access roles in the future.
         if (typeof roles === 'object' && roles instanceof Array) {
-          state.acronyms = roles.filter(role => !role.match(/offline_access|uma_authorization/));
+          state.acronyms = roles.filter(role => !role.match(/offline_access|uma_authorization|WEBADE_CFG_READ|WEBADE_CFG_READ_ALL/));
         } else {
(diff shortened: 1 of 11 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-84

frontend/src/store/modules/auth.js
```diff
@@ -20,8 +20,9 @@ export default {
         const payload = JSON.parse(atob(token.split('.')[1]));
         const roles = payload.realm_access.roles;
 
+        // TODO: this will require re-conceptualizing as acronyms will come from the DB, not the access roles in the future.
         if (typeof roles === 'object' && roles instanceof Array) {
-          state.acronyms = roles.filter(role => !role.match(/offline_access|uma_authorization/));
+          state.acronyms = roles.filter(role => !role.match(/offline_access|uma_authorization|WEBADE_CFG_READ|WEBADE_CFG_READ_ALL/));
```
parc-jason: should this be using utils.filterAppAcronyms?

loneil: I would, but this instance is in the frontend (and the utils class is in the backend).
Since this is the only instance in the frontend project that does this action I didn't refactor it out and duplicate the util function into the frontend.

parc-jason: Ahhh, yes.

PR #108, opened 2020-01-30
GETOK login ui change from UX feedback

TimCsaky:
# Description

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-663

- Changed the 'please log in' message on the home page to include a list of the common services and a login button. Will need further feedback on whether the wording is correct. wording on screenshot in ticket didn't quite make sense to me.
- updated the unit test that checks this text
- replaces string 'Natural Resources Get Token' with 'Common Service Get Token' in UI and readme's (further required changes seem beyond scope of this ticket)
- minor change to footer links styling

## Types of changes

## Checklist

## Further comments

some of the other unit tests relating to vuex and a regex were failing for me locally.
 I'm guessing that's expected or not requiring urgent attention (?)

This ticket included changing the title in the UI header from 'NR Get Token' to 'Common Service Get Token'. We may want to make this change at a system level in the environment variables and config maps (?)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-108

frontend/src/components/Home.vue
```diff
@@ -162,6 +186,10 @@ export default {
   },
   data() {
     return {
+      authRoutes: AuthRoutes,
+      KeycloakCommonServiceList: CommonServiceList.filter(function(x) {
```
jujaga: `CommonServiceList.filter(x => {`
This function doesn't depend on `this` scope so it can be declared anonymously

TimCsaky: good call.. .this view does look quite bland. apparently it will be getting more content soon.

frontend/src/components/Header.vue
```diff
@@ -12,7 +12,7 @@
       </a>
 
       <v-toolbar-title>
-        <v-btn class="title hidden-sm-and-down" color="text" text>{{ appTitle }}</v-btn>
+        <v-btn class="title hidden-sm-and-down" color="text" text>Common Service Get Token</v-btn>
```
loneil: We want this pulling from VUE_APP_TITLE, not hard coded. This allows us to control the title based on environment.

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-108

frontend/src/components/Header.vue
TimCsaky: totally agree. this was a temporary fix. i wondered if changing the environment variables would be done along with changes to server config map stuff in a different ticket ?

loneil: VUE_APP environment variables work differently. Look at the appTitle field and can see it's pulling from process.env.VUE_APP_TITLE, then see the .env files and the VUE_APP_* vars are set in there.
These could be config map-pulled values but for now are managed by the vue app variables.
For our discussions of changing how GETOK FE is hosted (served by it's api, rather then static on Caddy), this paradigm might change anyways.

But for current implementation should be able to just change the .env file for the title.

TimCsaky: awesome. that's good to know. thanks

frontend/src/components/Header.vue
```diff
@@ -69,6 +68,9 @@ export default {
   color: #fff;
   background: none;
 }
+.gov-header .v-btn.title{
+  text-transform: none;
```
loneil: Do we need to explicitly specify no text-transform? None is the default value. Is this needed to override something with a lower specificity?

frontend/src/components/Footer.vue
```diff
@@ -29,6 +30,9 @@ export default {};
   .footer .v-btn {
     color: #fff;
     text-decoration: none;
+    text-transform: capitalize;
+    font-size: 80%;
+    font-weight: normal;
```
loneil: Normal is the default font-weight, is this an override?

frontend/src/components/Header.vue
TimCsaky: yeah this is odd but i think v-btn default is uppercase. https://vuetifyjs.com/en/components/buttons

frontend/src/components/Footer.vue
TimCsaky: yeah this does seem unnecessary but i believe v-btn has a default font weight of 500. that's what the browser inspector is showing me. i think its in the base vuetify. 500 is a bit heavy if i am trying to match the style in ticket i think

frontend/src/components/Header.vue
loneil: Ah right, yeah that makes sense.
However, I think we'd want to stick with Veutify default styling if possible as it's an implementation of Material Design. Could check with Mina if it's important to deviate from the Veutify styles. If so we might want to look at setting global styles, as we want the app to be consistent. (ie, if this button is not capitalized, why are all the others in the app).

TimCsaky: agreed. i'm leaving it as default vuetify style. using a global stylesheet later sounds like a great idea

TimCsaky:
one unit test is failing:

FAIL tests/unit/components/WebAdeVisualizer/SecurityUtils.spec.js (5.224s)
SecurityUtils.vue
× does not search if the form is not valid (218ms)
Expected substring: "Enter a Regex to search on"
Received string: "<div class=\"container\">  <div><div cl.... (all the html for the component)

I'm struggling to fix this. i think i need access to a webade client to see the regex field?

loneil:
> one unit test is failing:
> 
> FAIL tests/unit/components/WebAdeVisualizer/SecurityUtils.spec.js (5.224s)
> SecurityUtils.vue
> × does not search if the form is not valid (218ms)
> Expected substring: "Enter a Regex to search on"
> Received string: "<div class="container">
> 
> <div cl.... (all the html for the component)
> I'm struggling to fix this. i think i need access to a webade client to see the regex field?

I'll take a look with you in a bit. Shouldn't relate to any access as this is just testing form validation and it's not related to any authentication state.

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-108

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-108

PR #110, opened 2020-02-06
New App registration reply email template

TimCsaky:
# Description
created an Outlook email template to be used by the team when replying to registration requests.
Template was created in a round-about way using IE 'send page as email' feature.

## Checklist


jujaga:
We should put this template file in the wiki instead as this template isn't directly leveraged by any part of the codebase at this time.

PR #116, opened 2020-02-11
DB queries for getting audit history by acronym

loneil:
# Description
Populate DB call for getting audit history.
This schema will change with new history setup after UX decisions, so not permanent, but structure in place now.

## Types of changes

## Checklist

### DIFF
--- a/backend/src/components/audit.js
+++ b/backend/src/components/audit.js
@@ -1,8 +1,8 @@
 const log = require('npmlog');
 
-// const {
-//   auditService
-// } = require('../services');
+const {
+  auditService
+} = require('../services');
 const utils = require('./utils');
 
 const audit = {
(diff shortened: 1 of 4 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-116

backend/src/components/audit.js
```diff
@@ -18,17 +18,25 @@ const audit = {
...
       if (history) {
-        return history;
+        const hsts = history.Lifecycles.map(lc => {
+          const hst = {};
+          hst.date = lc.createdAt;
+          hst.acronym = history.acronym;
+          hst.user = lc.LifecycleHistories[0].User.username;
+          hst.environment = lc.env;
+          hst.details = lc.appConfig;
+          return hst;
```
jujaga: As this is direct object assignments, it'd be better to write it as the following:
```
return {
  acronym: history.acronym,
  date: lc.createdAt,
  details: lc.appConfig,
  environment: lc.env,
  user: lc.LifecycleHistories[0].User.username
}
```
This is slightly more memory efficient as there's one less object constant being instantiated -> instead the object is just constructed as part of the anonymous function definition directly.

backend/src/services/audit.js
```diff
@@ -0,0 +1,24 @@
+const db = require('../models');
+
+module.exports = {
+  async findHistory(acronym) {
+    return await db.Acronym.findOne({
+      where: { acronym: acronym },
```
jujaga: Might not be a bad idea to have this where be after the include attribute (alphabetical ordering convention seen in other models)

backend/src/components/audit.js
```diff
@@ -1,8 +1,8 @@
 const log = require('npmlog');
 
-// const {
-//   auditService
-// } = require('../services');
+const {
+  auditService
+} = require('../services');
```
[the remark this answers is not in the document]
loneil: Fixed

backend/src/components/audit.js
loneil: Right, good call. Fixed

backend/src/services/audit.js
loneil: Yes, fixed

PR #117, opened 2020-02-11
Refactor backend to check user acronym permissions from DB

jujaga:
# Description

This PR changes user acronym permission checking behavior to check from the DB instead of from the JWT roles. This refactor is necessary in order to allow enable upcoming UI changes. The list of user acronyms is jury-rigged into the existing /auth/token endpoint for now, but will likely be migrated to its own discrete endpoint at a later date.

[SHOWCASE-714](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-714)

## Types of changes

## Checklist

### DIFF
--- a/backend/src/components/appConfig.js
+++ b/backend/src/components/appConfig.js
@@ -2,10 +2,7 @@ const axios = require('axios');
 const config = require('config');
 const log = require('npmlog');
 
-const {
-  acronymService,
-  lifecycleService
-} = require('../services');
+const { acronymService, lifecycleService } = require('../services');
 const utils = require('./utils');
 const permissionHelper = require('./permissionHelpers');
 
@@ -168,7 +165,7 @@ const appConfig = {
           'Content-Type': 'application/json; charset=utf-8'
         }
       });
-      log.verbose('postAppConfig', utils.prettyStringify(webAdeResponse.data));
+      log.verbose('postAppConfig', JSON.stringify(webAdeResponse.data));
 
       await lifecycleService.create(configForm.applicationAcronym, generatedConfig.webAdeCfg, webadeEnv, userId);
 
@@ -217,10 +214,11 @@ const appConfig = {
   },
 
   /**
-   * Fetch all application configurations from webade.
-   * @param {string} webadeEnv - Which ISSS webade env.
+   *  @function getAppConfigs
+   *  Fetch all application configurations from webade.
+   *  @param {string} webadeEnv - Which ISSS webade env.
    */
-  getAppConfigs: async (webadeEnv) => {
+  getAppConfigs: async webadeEnv => {
     const endpoint = config.get(`serviceClient.getok${utils.toPascalCase(webadeEnv)}.endpoint`);
     const username = config.get(`serviceClient.getok${utils.toPascalCase(webadeEnv)}.username`);
     const password = config.get(`serviceClient.getok${utils.toPascalCase(webadeEnv)}.password`);
@@ -235,7 +233,6 @@ const appConfig = {
     const path = '/applicationConfigurations';
 
     try {
-
       // Get the app configurations array
       const webAdeUrl = endpoint + path;
       const webAdeResponse = await axios.get(webAdeUrl, {
@@ -244,7 +241,7 @@ const appConfig = {
           'Content-Type': 'application/json; charset=utf-8'
         }
       });
-      log.verbose('getAppConfigs', utils.prettyStringify(webAdeResponse.data));
+      log.verbose('getAppConfigs', JSON.stringify(webAdeResponse.data));
       return webAdeResponse.data;
     } catch (error) {
       log.error('getAppConfigs', error.message);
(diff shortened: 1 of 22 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-117

backend/src/components/appConfig.js
```diff
@@ -168,7 +165,7 @@ const appConfig = {
           'Content-Type': 'application/json; charset=utf-8'
         }
       });
-      log.verbose('postAppConfig', utils.prettyStringify(webAdeResponse.data));
+      log.verbose('postAppConfig', JSON.stringify(webAdeResponse.data));
```
parc-jason: I see that most of the calls to utils.prettyStringify have been changed to JSON.stringify.  Let's remove that method, update any of the remaining 3 calls to use JSON.stringify, and we can just pass in the indent there if required.  It's one fewer thing to test, one fewer proprietary call for readers to look up.

backend/src/components/auth.js
```diff
@@ -4,14 +4,26 @@ const jsonwebtoken = require('jsonwebtoken');
...
+  /**
+   *  @function getUserAcronyms
+   *  @param {object} req An Express request object
+   *  @param {object} _res An Express response object
+   *  @param {function} next A callback function
+   */
+  async getUserAcronyms(req, _res, next) {
+    if (req.user && !auth.isTokenExpired(req.user.jwt)) {
+      const acronyms = await userService.getUserAcronymList(req.user.id);
+      req.user.acronyms = Array.isArray(acronyms) ? acronyms : [];
```
parc-jason: Where is this req.user.acronyms used? Still looks like all the work is from permissionHelper.checkAcronymPermission?

backend/src/routes/auth.js
```diff
@@ -60,7 +54,7 @@ router.post('/refresh', [
   return res.status(200).json(refresh);
 });
 
-router.use('/token', auth.refreshJWT, auth.updateDBFromToken, async (req, res) => {
+router.use('/token', auth.refreshJWT, auth.updateDBFromToken, auth.getUserAcronyms, (req, res) => {
```
parc-jason: Don't see that auth.getUserAcronyms does anything of benefit.

jujaga: auth.getUserAcronyms main purpose is to insert an acronyms attribute holding an array of acronyms a user has access to into the user object response of /auth/token. This is needed for the front-end to still be able to load its state and know what acronyms a user has access to.

backend/src/components/permissionHelpers.js
```diff
@@ -1,55 +1,50 @@
 const log = require('npmlog');
-const utils = require('./utils');
 
+const { userService } = require('../services');
 
 const permissionHelpers = {
   // TODO: this is likely soon to be refactored out, as we will be pulling acronyms from the DB, not from access roles
```
loneil: Still need this?

backend/src/components/appConfig.js
jujaga: Fair enough at this point - there isn't as much value for us to preserve this util function at this point as it's pretty simple with what it's doing to begin with.

backend/src/components/permissionHelpers.js
```diff
@@ -69,7 +64,8 @@ const permissionHelpers = {
         return `Acronym '${acronymData.acronym}' is not permitted special access to NROS DMS`;
       }
     } catch (error) {
-      log.error(`Error occurred determening permission for WebADE POST. Error: ${error}`);
+      console.log(error);
```
loneil: console.log

backend/src/components/auth.js
jujaga: `getUserAcronyms` is used in the route layer to expose the list of acronyms a user is associated to the /auth/token response. This is needed for subsequent front-end work. The subtle difference with `permissionHelper.checkAcronymPermission` is that instead of offering an array of acronyms, it answers the question "given this token and the acronym you are asking for, are you allowed to do it", whereas `getUserAcronyms` answers the question "what acronyms does this user have access to?".

backend/src/components/permissionHelpers.js
jujaga: The `filterAppAcronymRoles` function is still being left in as it's a part of the pathway where we load acronym information from JWT into the database. We elected at this time to leave that pathway in as it'd be an easy interim solution to populate the DB with new acronyms/user associations through Keycloak instead of having to do a direct DB query to insert. The TODO comment I left in because at some point further down the line when we remove this pathway we'll want to remove this function in its entirety as well.

backend/src/components/permissionHelpers.js
jujaga: Good catch. Also lame that eslint no longer warns on this... fixed.

backend/src/components/permissionHelpers.js
loneil: Right, gotcha

PR #120, opened 2020-02-18
Administrative endpoint for registering users

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-96

Add a endpoint (temporary shim until we build an admin interface) that will register users to applcation acronyms.
If a acronym is not in the DB it will create it, same with user.
User must be exist in KC for this to work.

The common services team member doing this must have the new role GETOK_ADMIN_ADD_USER.

Use a GET (with a GETOK team user bearer token from the GETOK realm) on the following URL for this to work
/api/v1/acronyms/{ACRONYM}/addUser/{USERNAME}
username is llike "loneil@idir"

Best way to do this is to use the DEBUG Api Tester panel in GETOK (konami code to access)

## Types of changes

## Checklist

## Further comments

I added the functionality to get Keycloak users to the KeyCloakServiceClientManager class, but that is instantiated with client credentials auth information for a keycloak realm admin user. Realized we don't have config for this in GETOK for the GETOK realm, just for the YAMS realm.
Then realized that a **user** token can be used to fetch this, as long as it has the view-users scope, which our team members will have because we are realm admins. And this is a better way of controlling that access.
So to not have to add a whole bunch of new config/openshift setup/documentation for this other service client. I'm going with the user token access for getting users instead.
I left the methods in the KC manager classes as this may be useful in the common services YAMS realm at some point.

### DIFF
--- a/backend/src/components/acronyms.js
+++ b/backend/src/components/acronyms.js
@@ -22,6 +22,18 @@ const acronyms = {
       log.error('getAcronym', error.message);
       throw new Error(`An error occured fetching acronym details from GETOK database. ${error.message}`);
     }
+  },
+
+  registerUserToAcronym: async (acronym, username) => {
+    await acronymService.findOrCreateList(acronymList);
+
+    // Add user if they don't already exist
+    await userService.findOrCreate(req.user.id, req.user.displayName, req.user._json.preferred_username);
+
+    // Add update user-acronym association from JWT roles
+    await acronymList.forEach(value => {
+      userService.addAcronym(req.user.id, value);
+    });
   }
 };
 
--- a/backend/src/components/acronyms.js
+++ b/backend/src/components/acronyms.js
@@ -1,6 +1,7 @@
+const axios = require('axios');
 const log = require('npmlog');
 
-const { acronymService } = require('../services');
+const { acronymService, userService } = require('../services');
 
 const acronyms = {
   /**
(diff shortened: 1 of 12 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-120

backend/src/routes/v1/acronyms.js
```diff
@@ -31,4 +31,39 @@ acronyms.get('/:appAcronym', [
...
+  if (!req.params.appAcronym || !req.params.username) {
+    res.status(400).json({
+      message: 'Must supply app acronym and user (ex: myname@idir)'
+    });
+    return res;
+  }
+
+  try {
+    const token = req.headers.authorization.split(' ')[1];
+    const response = await acronymComponent.registerUserToAcronym(token, req.user.jwt.iss, req.params.appAcronym.toUpperCase(), req.params.username.toLowerCase());
```
jujaga: Wondering if we should embed the toUpperCase and toLowerCase directly into that component function instead. It'd make the function be more of an enforcer.

backend/src/components/realmAdminSvc.js
```diff
@@ -388,6 +387,23 @@ class RealmAdminService {
     //204 created...
     return response;
   }
+
+  // Get from the users list with the username query param. If we want to be able to filter by more can refactor this method
+  // to take in search parameters. https://www.keycloak.org/docs-api/5.0/rest-api/index.html#_users_resource
```
parc-jason: yeah, we can think more on this later; this is a good one for our immediate needs.
totally possible to call with the search parameter that will query against username, first or last name, or email.  We would then need to iterate over the results to ensure that there is a correct match from idir.  Would be good for accounts like mine that are initially done via github, then have an idir account associated with them.  But let's leave that when we build out a more robust admin interface.

`https://sso-dev.pathfinder.gov.bc.ca/auth/admin/realms/jbd6rnxw/users?briefRepresentation=false&first=0&max=100&search=sherman`

Then look in the federatedIdentities / identityProvider = idir / username =

backend/src/routes/v1/acronyms.js
parc-jason: saw jujaga's comment about moving uppercase and lowercase into the registerUserToAcronym function - would be a good idea.

backend/src/components/realmAdminSvc.js
loneil: Sounds good

backend/src/routes/v1/acronyms.js
loneil: Fixed

loneil: Fixed

backend/src/components/acronyms.js
```diff
@@ -22,6 +23,39 @@ const acronyms = {
...
       throw new Error(`An error occured fetching acronym details from GETOK database. ${error.message}`);
     }
+  },
+
+  registerUserToAcronym: async (token, kcRealm, acronym, username) => {
+    await acronymService.findOrCreateList([acronym]);
+
+    // Get user details from KC
+    const url = `${kcRealm.replace('realms', 'admin/realms')}/users?username=${username}`;
+    log.debug(url);
```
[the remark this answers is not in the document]
loneil: Fixed

PR #126, opened 2020-03-06
Refactor homepage

TimCsaky:
# Description

supports: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-748
mockup: https://www.figma.com/proto/PqqxnXfahmDGJYC74fj1bQ/onboarding_v5?scaling=min-zoom&node-id=30%3A60

-added a 'welcome' component to show on the home page
-the header, footer, nav were done as part of the skeleton app.
-contains a list of common services that is pulled from a utils/commonServices.js object.

## Types of changes

## Checklist

- [ ] I have checked that unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/app/frontend/.env
+++ b/app/frontend/.env
@@ -1,1 +1,1 @@
-VUE_APP_TITLE=Common Service
+VUE_APP_TITLE=Common Services
--- a/app/frontend/.env
+++ b/app/frontend/.env
@@ -1,1 +1,1 @@
-VUE_APP_TITLE=Common Service Get Token
+VUE_APP_TITLE=Common Service
--- a/app/frontend/src/components/Welcome.vue
+++ b/app/frontend/src/components/Welcome.vue
@@ -18,13 +18,13 @@
                 <ul>
                   <li>
                     <strong>New User:</strong> Go to
-                    <router-link to="/">"Request Account"</router-link> and submit the form to add your application.
+                    <router-link :to="{ name: 'About' }">"Request Account"</router-link> and submit the form to add your application.
                     <br />* Check your email for the confirmation, then you’re ready to go!
                   </li>
                   <li>
                     <strong>Existing User:</strong> Manage your application from
-                    <router-link to="/">"My Applications"</router-link> or
-                    <router-link to="/">"Request Account"</router-link> for a new application
+                    <router-link :to="{ name: 'MyApps' }">"My Applications"</router-link> or
+                    <router-link :to="{ name: 'About' }">"Request Account"</router-link> for a new application
                   </li>
                   <li>
                     <strong>Get Token for your application:</strong> follow the order of
@@ -60,7 +60,7 @@
               <div class="pa-2 py-4">
                 <ul>
                   <li
-                    v-for="(item) in KeycloakCommonServiceList"
+                    v-for="item in KeycloakCommonServiceList"
                     :key="item.name"
                   >{{item.abbreviation.toUpperCase()}} - ({{item.name}})</li>
                 </ul>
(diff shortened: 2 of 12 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-126

app/frontend/src/components/Welcome.vue
```diff
@@ -0,0 +1,98 @@
+<template>
```
loneil: Think this page is called "About" in the latest revisions
https://www.figma.com/proto/PqqxnXfahmDGJYC74fj1bQ/onboarding_v5?node-id=30%3A60&scaling=min-zoom
Maybe check with Mina and see if that is looking final then rename.

app/frontend/src/components/Welcome.vue
```diff
@@ -0,0 +1,98 @@
...
+    return {
+      KeycloakCommonServiceList: CommonServiceList.filter(
+        x => x.type === 'keycloak'
+      )
+    };
+  }
+};
+</script>
+
+<style>
```
loneil: Should be scoped.
The **ui li** margin setting here looks like it's thrown off the nav styling.

app/frontend/src/components/Welcome.vue
```diff
@@ -0,0 +1,98 @@
...
+
+<script>
+import { CommonServiceList } from '@/utils/commonServices.js';
+
+export default {
+  name: 'homepage',
+  data() {
+    return {
+      KeycloakCommonServiceList: CommonServiceList.filter(
+        x => x.type === 'keycloak'
```
loneil: Can refer to the enumerated constant CommonServiceTypes.KEYCLOAK here instead of hard coded string.

app/frontend/.env
```diff
@@ -1 +1 @@
-VUE_APP_TITLE=Common Service Get Token
+VUE_APP_TITLE=Common Service
```
jujaga: Common Service**s** (missing an S on Service)

app/frontend/src/components/Welcome.vue
```diff
@@ -0,0 +1,98 @@
...
+};
+</script>
+
+<style>
+.v-card {
+  margin-bottom: 30px;
+}
+ul li {
+  margin-bottom: 10px;
+}
```
jujaga: These should be redefined to equivalent rem or em units instead of px.

app/frontend/src/components/Welcome.vue
TimCsaky: good idea!

app/frontend/src/components/Welcome.vue
```diff
@@ -0,0 +1,98 @@
+<template>
+  <v-container>
+    <v-row class>
```
loneil: Should try and set this up without nested grids, as nothing floats next to each other.

TimCsaky: thanks. please see updated code

app/frontend/src/components/Welcome.vue
```diff
@@ -90,9 +75,9 @@ export default {
 
 <style scoped>
 .v-card {
-  margin-bottom: 2rem;
+  margin-bottom: 3rem;
```
jujaga: Now that I think about it, I suppose this margin is mainly to pad for spacing between cards. However, I think there's also an app-wide issue with the footer bar where default content will get masked by it. We can revisit it later, but if this margin is to try and address the footer overflow, then we should be adding margin to a higher level component or even the application level itself instead of doing it in the card.

PR #132, opened 2020-03-10
Ui for application cards

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-765

Style the new cards and add them in a foreach loop.
Still dummy data. Merging this in in the interest of having small manageable PRs.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/base/BaseActionCard.vue
+++ b/app/frontend/src/components/base/BaseActionCard.vue
@@ -1,9 +1,16 @@
 <template>
-  <router-link class="action-link" :to="{ name: linkName, params: linkParams }">
-    <v-card class="action-card">
-      <slot />
-    </v-card>
-  </router-link>
+  <v-hover>
+    <template v-slot="{ hover }">
+      <router-link class="action-link" :to="{ name: linkName, params: linkParams }">
+        <v-card tile :elevation="hover ? 12 : 2" class="action-card mx-auto">
+          <v-system-bar />
+          <div class="px-5 py-3">
+            <slot />
+          </div>
+        </v-card>
+      </router-link>
+    </template>
+  </v-hover>
 </template>
 
 <script>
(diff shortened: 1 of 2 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-132

app/frontend/src/components/base/BaseActionCard.vue
```diff
@@ -1,9 +1,16 @@
...
-    <v-card class="action-card">
-      <slot />
-    </v-card>
-  </router-link>
+  <v-hover>
+    <template v-slot="{ hover }">
+      <router-link class="action-link" :to="{ name: linkName, params: linkParams }">
+        <v-card tile :elevation="hover ? 12 : 2" class="action-card mx-auto">
+          <v-system-bar />
+          <div class="px-5 py-3">
```
jujaga: For a base component do you think we should be forcing these types of values in? Would be good to try to minimize the amount of "opinionated" stuff in base components so that things that leverage or derive off of it aren't forced.

app/frontend/src/components/base/BaseActionCard.vue
```diff
@@ -17,8 +24,15 @@ export default {
...
 <style scoped>
-.v-card .action-card {
-  background-color: #003366;
-  border-top: 10px;
+.action-card {
+  height: 100%;
+}
+
+.v-system-bar {
+  background-color: #38598a;
```
jujaga: Maybe we should consider having this come into baseactioncard as a prop instead of embedded in here? Thoughts?

loneil: Hmm yeah that's a good idea. Will leave that until next usage of BaseActionCard.

PR #139, opened 2020-03-13
Request Account form - basic setup

TimCsaky:
# Description
supports: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-764

- new requestAccount view.
- new component for RequestForm
- new /requestAccount route
- updated nav links to this page
- basic unit test checking form renders

## Types of changes

## Checklist

## Further comments

Obviously a work in progress.
Some issues i need to figure out:
- the form doesnt seem to load when page is refreshed, only when reached via clicking on a link.
- i've added a global stylesheet, loaded via main.js. for discussion i assume
- needs to implement api service layer


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-139

loneil:
@TimCsaky if you navigate to the home page (a brand new session) and then just click the Request Account nav link, it logs you in, but doesn't load the page properly.

app/frontend/src/assets/scss/style.scss
```diff
@@ -0,0 +1,35 @@
...
+.v-btn:hover {
+  text-decoration: underline;
+  opacity: 0.8;
+}
+
+.BC-Gov-PrimaryButton.light.v-btn {
+  background-color: #fff;
+  border: 2px solid #003366;
+  color: #003366 !important;
+}
```
jujaga: We need to have a discussion about how we want to standardize theming - I don't know if this file is the appropriate way to do it just yet.

app/frontend/src/components/base/BaseDialog.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <v-dialog v-model="show" width="500">
```
jujaga: With a width of 500, have we tested this on very small screens to see if there's any potential overflow issues?

app/frontend/src/components/RequestForm.vue
```diff
@@ -0,0 +1,199 @@
...
+      comments: '',
+      errorOccurred: false,
+      fieldValidations: FieldValidations,
+      registrationDialog: false,
+      registerSuccess: false,
+      valid: false,
+      userInfo: {
+        emailAddress: Vue.prototype.$keycloak.tokenParsed.email,
+        idir:  Vue.prototype.$keycloak.tokenParsed.preferred_username,
+        name:  Vue.prototype.$keycloak.tokenParsed.name
```
jujaga: Extra space between : and Vue

Optional: Also may want to consider creating a "form" top level object and having all form field content reside under that object instead. It makes it easier to see at a glance what pieces of data are form content, and what other pieces are for visual state.

app/frontend/src/components/RequestForm.vue
```diff
@@ -0,0 +1,199 @@
...
+          <label>E-mail</label>
+          <v-text-field
+            readonly
+            hide-details="auto"
+            solo
+            dense
+            single-line
+            outlined
+            flat
+            :value="userInfo.emailAddress"
```
jujaga: Wondering if this should be bound as a v-model to the respective values instead of just value - even if the fields are read-only.

app/frontend/src/components/RequestForm.vue
```diff
@@ -0,0 +1,199 @@
...
+  </v-container>
+</template>
+
+<script>
+import Vue from 'vue';
+import { FieldValidations } from '@/utils/constants.js';
+
+export default {
+  name: 'RequestForm',
+  data: function() {
```
jujaga: Optional style: Could be `data() {` and would still be equivalent.

app/frontend/src/components/RequestForm.vue
```diff
@@ -0,0 +1,199 @@
...
+      valid: false,
+      userInfo: {
+        emailAddress: Vue.prototype.$keycloak.tokenParsed.email,
+        idir:  Vue.prototype.$keycloak.tokenParsed.preferred_username,
+        name:  Vue.prototype.$keycloak.tokenParsed.name
+      }
+    };
+  },
+  computed: {
+  },
```
jujaga: If this isn't being used, this shouldn't be here.

app/frontend/src/components/RequestForm.vue
```diff
@@ -0,0 +1,199 @@
...
+          v.length <= FieldValidations.ACRONYM_MAX_LENGTH ||
+          `Acronym must be ${FieldValidations.ACRONYM_MAX_LENGTH} characters or less`,
+        v =>
+          /^(?:[A-Z]{2,}[_]?)+[A-Z]{1,}$/g.test(v) ||
+          'Incorrect format. Hover over ? for details.'
+      ],
+      comments: '',
+      errorOccurred: false,
+      fieldValidations: FieldValidations,
+      registrationDialog: false,
```
jujaga: What is `registrationDialog` used for? All three references have it set to false but it appears like nothing is actually leveraging it?

app/frontend/src/components/RequestForm.vue
```diff
@@ -0,0 +1,199 @@
...
+          //   comments: this.comments,
+          //   from: this.userInfo.emailAddress,
+          //   idir: this.userInfo.idir
+          // });
+
+          if (response) {
+            this.registrationDialog = false;
+            this.registerSuccess = true;
+          } else {
+            this.errorOccurred = true;
```
jujaga: Two options here:
1) (Keep try-catch) just throw a new error so it cascades to Line 186 - don't need to duplicate "error cases" when they do the same thing
2) (Drop try-catch) Change to Promise.then.catch pattern and wrap as needed (the API service will be returning raw axios responses, which are promise based, so we can leverage that instead of having to use our own try catch)

app/frontend/src/components/RequestForm.vue
```diff
@@ -0,0 +1,199 @@
...
+        }
+      }
+    }
+  }
+};
+</script>
+
+<style scoped>
+.request-form {
+  max-width: 450px;
```
jujaga: Change this to rem/em units if possible/applicable.

app/frontend/src/components/base/BaseDialog.vue
```diff
@@ -0,0 +1,57 @@
...
+<script>
+import '@/assets/scss/style.scss';
+
+export default {
+  name: 'BaseDialog',
+  props: {
+    'show': Boolean
+  },
+  methods: {
+    closeDialog (){
```
jujaga: Use VSCode autoformatter to fix this `closeDialog() {`

app/frontend/src/components/base/BaseDialog.vue
```diff
@@ -0,0 +1,57 @@
...
+      this.$emit('close-dialog');
+    }
+  }
+};
+</script>
+
+<style scoped>
+.v-card__text {
+  display: flex !important;
+  padding: 20px;
```
jujaga: Use rem/em units if possible

app/frontend/src/components/base/BaseDialog.vue
```diff
@@ -0,0 +1,57 @@
...
+}
+.dialog-icon {
+  margin-right: 10px;
```
jujaga: Use rem/em units if possible

app/frontend/src/components/bcgov/BCGovFooter.vue
```diff
@@ -26,14 +26,13 @@ export default {
 
 <style scoped>
 .gov-footer.v-footer {
-  bottom: 0;
+  flex-shrink: 0;
```
jujaga: This should be under border-top (alphabetical ordering)

app/frontend/src/router/index.js
```diff
@@ -40,6 +40,14 @@ const router = new VueRouter({
       },
       props: true
     },
+    {
+      path: '/apps/requestAccount',
```
jujaga: This should just be `/requestAccount`. This isn't part of apps flow.

app/frontend/src/utils/constants.js
```diff
@@ -0,0 +1,9 @@
+export const FieldValidations = Object.freeze({
+  ACRONYM_MAX_LENGTH: 30,
+  ACRONYM_MIN_LENGTH: 3,
+  NAME_MAX_LENGTH: 120,
```
jujaga: Let's have this value be in alphabetical order

app/frontend/src/views/RequestAccount.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+  <v-container>
+    <BaseSecure>Request Account</BaseSecure>
+    <RequestForm />
```
jujaga: BaseSecure needs to wrap everything - the way it is right now you can get content leakage. This may be related to the comment from @loneil 
```
<BaseSecure>
  <div>Request Account</div>
  <RequestForm />
</BaseSecure>
```

app/frontend/src/assets/scss/style.scss
TimCsaky: agreed. but i'm struggling to find a better way right now.

app/frontend/src/components/base/BaseDialog.vue
TimCsaky: good idea. this should be max-width="500"

app/frontend/src/components/RequestForm.vue
TimCsaky: yeah sounds like a good idea

app/frontend/src/components/bcgov/BCGovFooter.vue
TimCsaky: i'm trying this

app/frontend/src/components/RequestForm.vue
TimCsaky: oh yes. cheers

app/frontend/src/router/index.js
TimCsaky: good idea

app/frontend/src/views/RequestAccount.vue
TimCsaky: that was the reason. thanks!

app/frontend/src/components/RequestForm.vue
TimCsaky: done

app/frontend/src/components/RequestForm.vue
TimCsaky: good catch. before the form was in a dialog box

app/frontend/src/components/RequestForm.vue
TimCsaky: ok. i think i have it corrected

app/frontend/src/components/base/BaseDialog.vue
TimCsaky: done

app/frontend/src/assets/scss/style.scss
TimCsaky: i've created a separate jira ticket to work on this.

app/frontend/src/assets/scss/style.scss
```diff
@@ -0,0 +1,35 @@
...
+body > .v-content {
+  flex: 1 0 auto;
+}
+
+.BC-Gov-PrimaryButton.v-btn {
+  background-color: #003366;
+  border: none;
+  border-radius: 4px;
+  color: white !important;
```
loneil: Would be best to stay away from !important as much as possible (though Vuetify can make that tricky).
We can re-assess when we make some theming decisions

TimCsaky: ok

PR #142, opened 2020-03-17
Display service client status (with dummy data)

loneil:
# Description
Put in the structure to display KeyCloak client status
This uses hard coded dummy data to get the spec and scaffold and UI in. Next PR will include backend impl and UTs.

Hard coded to only have (fake) client data for WORG and MSSC in this impl.

Can just review code and test it out fuller on the next PR.

## Checklist

### DIFF
--- a/app/frontend/src/components/ApplicationList.vue
+++ b/app/frontend/src/components/ApplicationList.vue
@@ -31,9 +31,18 @@
                   transition="scale-transition"
                 >
                   <p>
-                    Dev: Not
-                    <br />Test: real
-                    <br />Prod: data
+                    Dev:
+                    <span
+                      :class="acr.devStatus ? 'green--text' : ''"
+                    >{{acr.devStatus ? clientExistsText : clientNotExistsText}}</span>
+                    <br />Test:
+                    <span
+                      :class="acr.testStatus ? 'green--text' : ''"
+                    >{{acr.testStatus ? clientExistsText : clientNotExistsText}}</span>
+                    <br />Prod:
+                    <span
+                      :class="acr.prodStatus ? 'green--text' : ''"
+                    >{{acr.prodStatus ? clientExistsText : clientNotExistsText}}</span>
                   </p>
                 </v-skeleton-loader>
               </div>
@@ -52,18 +61,23 @@ export default {
   name: 'ApplicationList',
   data: () => ({
     // TEMP for testing skeletons
-    waiting: true
+    waiting: true,
+    clientExistsText: 'Available',
+    clientNotExistsText: 'Not Available'
   }),
   computed: {
     ...mapGetters('user', ['acronyms'])
   },
   methods: {
-    ...mapActions('user', ['getUserAcronyms'])
+    ...mapActions('user', ['getUserAcronyms', 'fillInAcronymClientStatus'])
   },
-  mounted() {
-    this.getUserAcronyms();
-    // TEMP for testing skeletons
-    setTimeout(() => (this.waiting = false), 1000);
+  async mounted() {
+    //TODO: consider moving getUserAcronyms to on login instead of this component
+    await this.getUserAcronyms();
+
+    this.waiting = true;
+    await this.fillInAcronymClientStatus();
+    this.waiting = false;
   }
 };
 </script>
--- a/app/frontend/src/components/ApplicationList.vue
+++ b/app/frontend/src/components/ApplicationList.vue
@@ -30,20 +30,7 @@
                   :loading="waiting"
                   transition="scale-transition"
                 >
-                  <p>
-                    Dev:
-                    <span
-                      :class="acr.devStatus ? 'green--text' : ''"
-                    >{{acr.devStatus ? clientExistsText : clientNotExistsText}}</span>
-                    <br />Test:
-                    <span
-                      :class="acr.testStatus ? 'green--text' : ''"
-                    >{{acr.testStatus ? clientExistsText : clientNotExistsText}}</span>
-                    <br />Prod:
-                    <span
-                      :class="acr.prodStatus ? 'green--text' : ''"
-                    >{{acr.prodStatus ? clientExistsText : clientNotExistsText}}</span>
-                  </p>
+                  <p v-html="setClientTexts(acr)" />
                 </v-skeleton-loader>
               </div>
             </div>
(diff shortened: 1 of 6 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-142

app/frontend/src/components/ApplicationList.vue
```diff
@@ -52,18 +61,23 @@ export default {
   name: 'ApplicationList',
   data: () => ({
     // TEMP for testing skeletons
-    waiting: true
+    waiting: true,
+    clientExistsText: 'Available',
+    clientNotExistsText: 'Not Available'
```
jujaga: Extending on the above comment... we could consider just making this a computed instead and pass in a boolean. Data usually implies it needs to be permuted, but the information stored in those fields are not.

app/frontend/src/components/ApplicationList.vue
```diff
@@ -52,18 +61,23 @@ export default {
...
   methods: {
-    ...mapActions('user', ['getUserAcronyms'])
+    ...mapActions('user', ['getUserAcronyms', 'fillInAcronymClientStatus'])
   },
-  mounted() {
-    this.getUserAcronyms();
-    // TEMP for testing skeletons
-    setTimeout(() => (this.waiting = false), 1000);
+  async mounted() {
+    //TODO: consider moving getUserAcronyms to on login instead of this component
```
jujaga: If it's done on "login", that means this needs to be executed on the App.vue level or even higher up (main.js). There's benefits and drawbacks of making the call earlier in the lifecycle.

app/frontend/src/store/modules/user.js
```diff
@@ -37,6 +53,21 @@ export default {
...
+    },
+    /**
+     * @function fillInAcronymClientStatus
+     * Fetch the service clients for the acronyms the user has
+     * @param {object} context The store context
+     */
+    async fillInAcronymClientStatus(context) {
+      try {
+        const response = await UserService.getServiceClients(
+          Vue.prototype.$keycloak.subject
```
jujaga: This may not be safe to call yet without the prerequisite keycloak ready and authenticated checks. Will have to investigate with the authentication abstraction refactor.

app/src/routes/v1/users.js
```diff
@@ -28,4 +28,28 @@ usersRouter.get('/:keycloakId/acronyms', [
   }
 });
 
+/** Returns acronym service clients associated with a user */
+usersRouter.get('/:keycloakId/acronyms/clients', [
+  param('keycloakId', 'Must be a valid UUID')
+    .exists().custom(value => validator.isUUID(value))
```
jujaga: We actually don't want the .exists() as it will cause a double report in the errors array.

app/frontend/src/components/ApplicationList.vue
loneil: Yes that's one of the discussion points for tomorrow.

app/frontend/src/store/modules/user.js
loneil: Indeed, didn't want to do that multi-property check. We'll refactor it a bit after discussion tomorrow.

app/frontend/src/components/ApplicationList.vue
```diff
@@ -31,9 +31,18 @@
...
+                      :class="acr.devStatus ? 'green--text' : ''"
+                    >{{acr.devStatus ? clientExistsText : clientNotExistsText}}</span>
+                    <br />Test:
+                    <span
+                      :class="acr.testStatus ? 'green--text' : ''"
+                    >{{acr.testStatus ? clientExistsText : clientNotExistsText}}</span>
+                    <br />Prod:
+                    <span
+                      :class="acr.prodStatus ? 'green--text' : ''"
+                    >{{acr.prodStatus ? clientExistsText : clientNotExistsText}}</span>
```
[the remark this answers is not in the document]
loneil: Had the same thought. Thanks for confirming, will do.

app/frontend/src/components/ApplicationList.vue
loneil: I'll see how it shakes out with refactoring to a method. This text is specific to how the component displays the boolean condition from the store's object, so I considered it component-specific data...
But may change after the above

app/src/routes/v1/users.js
loneil: kk

app/frontend/src/store/modules/user.js
loneil: Will leave this and revisit with Jeremy after auth state changes.

app/frontend/src/components/ApplicationList.vue
loneil: Will leave this and revisit with Jeremy after auth state changes.

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-142

app/frontend/src/store/modules/user.js
```diff
@@ -14,6 +14,22 @@ export default {
...
         state.acronyms = acronyms;
       }
+    },
+    setAcronymClientStatus: (state, clientsForUser) => {
+      if (Array.isArray(clientsForUser) && clientsForUser.length) {
+        state.acronyms.forEach(acr => {
+          const clientSet = clientsForUser.find(client => client.acronym === acr.acronym);
+          acr.devStatus = (clientSet && clientSet.dev && clientSet.dev.enabled) ? true : false;
+          acr.testStatus = (clientSet && clientSet.test && clientSet.test.enabled) ? true : false;
+          acr.prodStatus = (clientSet && clientSet.prod && clientSet.prod.enabled) ? true : false;
```
jujaga: All of these lines don't need a ternary - you can just do the following and it'll be the same:
`acr.devStatus = clientSet && clientSet.dev && clientSet.dev.enabled`

loneil: Yes right, had left that in from when I was doing text.

app/frontend/src/store/modules/user.js
jujaga: Will be using vuex lazy loading on view level to address on-demand loading.

app/frontend/src/components/ApplicationList.vue
jujaga: Will be using vuex lazy loading on view level to address on-demand loading.

PR #144, opened 2020-03-18
Bc fonts

TimCsaky:
# Description

Use BC fonts in frontend UI
supports ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-870

as discussed, i've implemented this with the bc-sans npm package.
if we dont like this, we can put the fonts in our public/assests

we might need to use !important to overide vuetify (sparingly)
alternative could be to add a custom class name (like .bcgov) to the body tag (in index.html) 
and then use inheritance.
i think vuetify does define the font on some of the headings. maybe we override these on an individual basis.

i also think the bcsans font doesnt look very smooth (on firefox anyway)

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/assets/scss/style.scss
+++ b/app/frontend/src/assets/scss/style.scss
@@ -1,3 +1,12 @@
+/* Typography */
+body .v-application {
+  font-family: -apple-system, BlinkMacSystemFont, 'BCSans', ‘Noto Sans’, Verdana, Arial, sans-serif !important;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  font-size: 1em;
+  line-height: 1.25;
+}
+
 /* sticky footer */
 html,
 body {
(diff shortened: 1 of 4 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-144

app/frontend/src/assets/scss/style.scss
```diff
@@ -1,3 +1,12 @@
+/* Typography */
+body .v-application {
+  font-family: -apple-system, BlinkMacSystemFont, 'BCSans', ‘Noto Sans’, Verdana, Arial, sans-serif !important;
```
jujaga: The quotes around Noto Sans looks a bit strange - is this intended?

TimCsaky: well spotted! i think the official way is double quotes around names made of more than one word. i'll change this

PR #145, opened 2020-03-19
fixed line height and headings inline with bcgov style guide

TimCsaky:
# Description

final change (for now) to typography to get headings and paragraphs to match bcgov design guidelines.
as seen here:
https://developer.gov.bc.ca/Design-System/Typography
(not the best looking website)

thought this was worth doing now as it will affect spacing of new components

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/public/index.html
+++ b/app/frontend/public/index.html
@@ -6,7 +6,7 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
     <title><%= process.env.VUE_APP_TITLE %></title>
-    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Noto+Sans:100,300,400,500,700,900">
+    <!-- <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Noto+Sans:100,300,400,500,700,900"> -->
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700,900">
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Material+Icons">
     <link rel="stylesheet" href="https://unpkg.com/@mdi/font@latest/css/materialdesignicons.min.css">
--- a/app/frontend/public/index.html
+++ b/app/frontend/public/index.html
@@ -6,6 +6,7 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
     <title><%= process.env.VUE_APP_TITLE %></title>
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Noto+Sans:100,300,400,500,700,900">
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700,900">
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Material+Icons">
     <link rel="stylesheet" href="https://unpkg.com/@mdi/font@latest/css/materialdesignicons.min.css">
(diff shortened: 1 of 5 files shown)


TimCsaky:
Let me know if you'd like me to include other changes requested by Mina on rocket chat yesterday in this pr

loneil:
> Let me know if you'd like me to include other changes requested by Mina on rocket chat yesterday in this pr

No, we'll wrap those up and other styling things we come along in another ticket. Let's get the public pages all in and then can have a closer review of styling with Mina for everything.

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-145

app/frontend/public/index.html
```diff
@@ -6,6 +6,7 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
     <title><%= process.env.VUE_APP_TITLE %></title>
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Noto+Sans:100,300,400,500,700,900">
```
TimCsaky: i'm ok if we dont include noto-sans font as well in the html header.
it's only used in the bcgov header. let me know if you'd rather take this out

loneil: Are noto-sans and roboto not used anywhere then? We can take them out to save the loading if so.

app/frontend/src/assets/scss/style.scss
```diff
@@ -1,12 +1,28 @@
...
+h1, h2, h3, h4, h5 {
+  line-height: 1.2 !important;
+}
+
+.v-application .gov-header .title {
+  @extend h2;
+  font-family: "Noto Sans", Verdana, Arial, sans-serif !important;
+}
+
+p{
```
loneil: missing a space between p and { (we likely don't have the lint rule for scss?)

app/frontend/src/assets/scss/style.scss
```diff
@@ -1,12 +1,28 @@
...
+  line-height: 1.5 !important;
 }
 
+h1 { font-size: 2.074em; /* (37.32px) */ }
+h2 { font-size: 1.728em; /* (31.10px) */ }
+h3 { font-size: 1.444em; /* (25.92px) */ }
+h4 { font-size: 1.2em; /* (21.60px) */ }
+
+h1, h2, h3, h4, h5 {
+  line-height: 1.2 !important;
```
loneil: Would really not rather throw !important around so much. But as we've discussed Vuetify has a high specificity due to theme classes. I think we can fight this with our own theme down the line.

app/frontend/public/index.html
TimCsaky: - i put in noto-sans for the .gov-header .title (as they do on the bcgov design site).
- roboto is the vuetify default. I wonder if we might want it for some 'native' vuetify components
i'm not sure what the best thing to do here is. as a compromise i could just remove the noto-sans font that is only used in the header. this looked ok before in roboto.

loneil: Ok sounds good then.

app/frontend/src/assets/scss/style.scss
TimCsaky: yeah i agree.. i struggled to find a better way and then read it was common practice for overriding vuetify. i think it's ok as long as you can override this as well. one alternative would be to add a class to the body tag (or some other high-level html tag and then override the specificity(?) with a rule like: body.bcgov h1{...} not sure if that works.

TimCsaky: let me try and get rid of those important tags now

loneil: Yeah I've had to add classes and do longer selectors to get the specificity up in the past, which is bad practice as well. No no winning strategy with Vuetify from what I've seen. I think it's ok to have the importants at this top level since it's defining it for the whole app, but again I think it's an issue we can kick down the road till we see what we can do with a theme.

TimCsaky: you were right.. there was no need for all those !importants

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-145

PR #146, opened 2020-03-20
Abstract Keycloak behind Vuex module layer

jujaga:
# Description

This PR refactors Keycloak so that it is abstracted away behind a Vuex layer. The Vuex store has also been reshaped so that modules are registered and lazy loaded as needed in order to better split apart webpack bundling. This improves overall separation of concerns, as well as improves our ability to unit test effectively.

**Other changes:**
- Change structure of unit tests to register stores as needed for mocks
- Remove unused secure view
- Change user module to register on myApps view, and dispatch loader on component mount

[SHOWCASE-872](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-872)

## Types of changes

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-146

app/frontend/src/store/modules/user.js
```diff
@@ -30,44 +31,58 @@ export default {
...
-              context.commit('setAcronyms', response.data);
-            }
-          })
-          .catch(error => {
-            // TODO: Create top-level global state error message
-            console.error(error); // eslint-disable-line no-console
-          });
+    async getAcronymClientStatus(context) {
+      try {
+        if (context.rootGetters['auth/authenticated']) {
```
parc-jason: I don't like this.  Is there any way to bury that in the service/class or some kind of abstraction?  Is  context.rootGetters['auth/subject'] used in more places than this?  I imagine context.rootGetters['auth/authenticated'] will be. Are we going to wrap all calls to services with that?

app/frontend/src/store/modules/auth.js
```diff
@@ -0,0 +1,19 @@
...
+
+export default {
+  namespaced: true,
+  state: {},
+  getters: {
+    authenticated: () => Vue.prototype.$keycloak.authenticated,
+    createLoginUrl: () => options => Vue.prototype.$keycloak.createLoginUrl(options),
+    createLogoutUrl: () => options => Vue.prototype.$keycloak.createLogoutUrl(options),
+    moduleLoaded: () => !!Vue.prototype.$keycloak,
+    ready: () => Vue.prototype.$keycloak.ready,
```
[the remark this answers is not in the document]
jujaga: Fair point on clarity. I've gone ahead and renamed them to `keycloakReady` and `keycloakSubject` to be more explicit.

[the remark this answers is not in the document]
parc-jason: maybe `authenticationReady` and `authenticationSubject` or `jwt`

app/frontend/src/store/modules/user.js
loneil: This is how vuex stores communicate with each other, I'm not sure there's a way around it. Maybe could extract to a utility method in the same store. If possible that would be better because hard coding the string isn't ideal if it's used more than once.
So this wouldn't be used in components, services, etc. Only in vuex modules themselves when they need to talk to a parent or sibling module

loneil: @jujaga look at the section "Accessing Global Assets in Namespaced Modules" here
https://vuex.vuejs.org/guide/modules.html
might be able to make the syntax a bit cleaner at least?

app/frontend/src/store/modules/user.js
```diff
@@ -16,19 +16,26 @@ export default {
...
+    setAcronymClientStatus: (state, userClients) => {
+      if (Array.isArray(userClients) && userClients.length) {
+        state.acronyms = state.acronyms.map(acr => {
+          const clientSet = userClients.find(client => client.acronym === acr.acronym);
+          acr.clientStatus = {
+            dev: clientSet && clientSet.dev && clientSet.dev.enabled,
+            test: clientSet && clientSet.test && clientSet.test.enabled,
+            prod: clientSet && clientSet.prod && clientSet.prod.enabled
+          };
+          return acr;
```
loneil: Awesome, much rather use a map than a foreach

PR #147, opened 2020-03-20
Refactor requestForm to use auth store and Documentation

jujaga:
# Description

This PR refactors the RequestForm component to use the auth vuex store layer instead of directly invoking the Keycloak plugin. This change was required in order to more cleanly do unit testing for any component which needs to ask about the current user.

**Other changes:**
- Add disabled button style as per design guidelines
- Fix typo in logout class name in BaseAuthButton
- Add some documentation on Vuex unit testing and considerations

[SHOWCASE-824](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-824)
[SHOWCASE-872](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-872)
[SHOWCASE-889](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-889)

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/RequestForm.vue
+++ b/app/frontend/src/components/RequestForm.vue
@@ -88,8 +88,8 @@
       </v-row>
     </v-form>
     <div class="justify-center pb-8">
-      <v-btn class="BC-Gov-PrimaryButton light float-left" text @click="cancel()">Cancel</v-btn>
-      <v-btn class="BC-Gov-PrimaryButton float-right" text @click="postRegistrationForm()">Submit</v-btn>
+      <v-btn class="request-form-cancel-btn BC-Gov-PrimaryButton light float-left" text @click="cancel()">Cancel</v-btn>
+      <v-btn class="request-form-submit-btn BC-Gov-PrimaryButton float-right" text @click="postRegistrationForm()">Submit</v-btn>
     </div>
 
     <BaseDialog v-bind:show="errorOccurred" @close-dialog="errorOccurred = false">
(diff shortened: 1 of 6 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-147

app/frontend/src/components/RequestForm.vue
```diff
@@ -170,14 +172,20 @@ export default {
...
+      this.form.idir = this.tokenParsed.preferred_username;
+    },
     resetState() {
       this.errorOccurred = false;
       this.registerSuccess = false;
     }
   },
   mounted() {
     this.resetState();
+    this.resetForm();
```
TimCsaky: looks good and all works. just wondered if doing resetForm() is the best way to load the data.form.idir and data.form.from. could be hard coded into the template(?) also you may not want to reset the form when a user navigates to another page and then back.

jujaga: User navigation reset is a minor issue as 2 of the 4 fields are read-only to the user, and the only "real" field we care about is the acronym. It's highly unlikely a user will navigate to a different page while working on this form, and even if they did, there really isn't that much for them to fill out anyways.

The reason the values are loaded on mount via a resetForm method is because this is the earliest in the component lifecycle that it can get the JWT information from the vuex auth layer.

PR #149, opened 2020-03-23
Onboarding Stats for Admin page

TimCsaky:
# Description

see: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-825

We want a list of all the service clients registered in GETOK and in which keycloak realms they exist

## Types of changes

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-149

app/frontend/src/router/index.js
```diff
@@ -40,6 +40,17 @@ const router = new VueRouter({
...
+    // tempoarary place for admin page
+    {
+      path: '/admin',
+      name: 'Admin',
+      component: () => import(/* webpackChunkName: "admin" */ '@/views/Admin.vue'),
+      meta: {
+        requiresAuth: true
+      },
+    },
+
```
jujaga: You will get merge conflicts with #150 . You'll have to drop this chunk from your git branch.

app/frontend/src/views/Admin.vue
```diff
@@ -0,0 +1,19 @@
...
+import Stats from '@/components/Stats.vue';
+
+export default {
+  name: 'Admin',
+
+  components: {
+    Stats
+  }
+};
+</script>
```
jujaga: You will get merge conflicts with #150 . You'll have to drop this chunk from your git branch.

app/src/routes/v1/keycloak.js
```diff
@@ -0,0 +1,42 @@
...
+], async (req, res) => {
+  // requires user role "GETOK_ADMIN"
+  if (!req.user.jwt.realm_access.roles.includes('GETOK_ADMIN')) {
+    // mock for now
+    // do we need a GETOK_ADMIN permission check?
+    const permissionErr = false;
+    if (permissionErr) {
+      return res.status(403).json({
+        message: permissionErr
+      });
```
jujaga: You'll want to update this to be using the api-problem path instead of directly responding.

app/src/routes/v1/keycloak.js
```diff
@@ -0,0 +1,42 @@
...
+    } else {
+      res.status(200).json(result);
+    }
+
+
+  } catch (error) {
+    log.error(error);
+    res.status(500).json({
+      message: error.message
+    });
```
jujaga: You'll want to update this to be using the api-problem path instead of directly responding.

app/src/docs/v1.api-spec.yaml
```diff
@@ -169,32 +169,19 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
-  "/keyCloak/clients/{keycloakEnv}/{acronym}":
+  "/keyCloak/clients":
```
jujaga: Let's change this to /serviceClients/clients as we want to 1) only focus on service clients (as per discussion) and 2) we want to try to avoid referring to keycloak as much as possible because keycloak is implementation specific.

app/src/components/stats.js
```diff
@@ -0,0 +1,95 @@
...
+
+
+
+  /**
+   * @function getClientsFromEnv
+   * Utility function to call the KC service to get service clients for each realm
+   * @param {string} kcEnv The KC env
+   * @returns {object[]} An array of service clients
+   */
+  getClientsFromEnv: async (kcEnv) => {
```
TimCsaky: This is similar to a function in the backend users component but does not filter by acronym

app/src/components/stats.js
```diff
@@ -0,0 +1,95 @@
...
+    const {
+      endpoint: realmBaseUrl,
+      username: clientId,
+      password: clientSecret,
+      realm: realmId
+    } = config.get(realmKey);
+
+    const realmSvc = new RealmAdminService({ realmBaseUrl, clientId, clientSecret, realmId });
+
+    const kcScMgr = new KeyCloakServiceClientManager(realmSvc);
```
TimCsaky: this is where I am seeing errors related to the realmAdminService.
by the way, for the service client inmy local.json config file I am using:
```
  "serviceClient": {
    "keycloak": {
      "dev": {
        "username": "KEYCLOAK_REALM_ADMIN",
        "password": "xxxx-xxx-xxxx-0359917c7f4a"
      },...
```
does that look right?

app/src/docs/v1.api-spec.yaml
```diff
@@ -169,32 +169,19 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
-  "/keyCloak/clients/{keycloakEnv}/{acronym}":
```
TimCsaky: I know this spec will need tidying up once we are sure about the name for this endpoint

app/src/routes/v1/stats.js
```diff
@@ -0,0 +1,29 @@
+const statsRouter = require('express').Router();
+const Problem = require('api-problem');
+//const log = require('npmlog');
+
+const stats = require('../../components/stats');
+
+// fetches all the service clients for all acronyms in all KC realms
+statsRouter.get('/serviceClients', async (req, res) => {
+
+  // requires user role "GETOK_ADMIN"
```
TimCsaky: i see the bearer token in the header of the request (req) but i still havent found a good way to check the current users keycloak role :(

jujaga: It could look something like this:
`statsRouter.get('/serviceClients', keycloak.protect('realm:GETOK_ADMIN'), async (req, res) => {`
Examples from their docs: https://github.com/keycloak/keycloak-quickstarts/blob/latest/service-nodejs/app.js

app/frontend/src/components/Clients.vue
```diff
@@ -0,0 +1,113 @@
...
+      this.loading = false;
+    }
+  },
+  methods: {
+    // get table data from frontend service layer
+    getData() {
+      keycloakService.getServiceClients().then(response => {
+        if (response) {
+
+          // reformat serviceClients data into array of objects that will work for the data table
```
TimCsaky: I spent way too long trying to reformat the data using other array functions. this approach might be a bit crude but it's the only way i could get it working

app/frontend/src/components/Clients.vue
```diff
@@ -0,0 +1,113 @@
...
+      });
+    }
+  },
+  mounted() {
+    this.getData();
+  }
+};
+</script>
+
+<style scoped>
```
TimCsaky: i'll remove this

app/frontend/src/components/Clients.vue
```diff
@@ -0,0 +1,113 @@
...
+.kc-table .active {
+  width: 10px;
+  height: 10px;
+  display: block;
+  background: blue;
+}
+</style>
```
TimCsaky: i'm going to remove this in my next commit once i've figured out the best way to rebase from the remote feature-refactor branch

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-149

app/frontend/src/components/Stats.vue
```diff
@@ -1,79 +1,113 @@
...
-        .getServiceClients()
-        .then(response => {
-          if (response) {
-            return response;
+      keycloakService.getServiceClients().then(response => {
+        if (response) {
+
+          // reformat serviceClients data into array of objects that will work for the data table
+          var output = [];
+          var tmpArray = [];
```
jujaga: Always use `const` if possible, or `let` if  the variable needs to be reassigned.
You likely don't need both output and tmpArray - just pick one of them and stick with it.

app/frontend/src/components/Stats.vue
```diff
@@ -1,79 +1,113 @@
...
+            // update object
+            tmpArray[sc.clientId].name = sc.clientId;
+            if (sc.realm === 'dev') tmpArray[sc.clientId].dev = 1;
+            if (sc.realm === 'test') tmpArray[sc.clientId].test = 1;
+            if (sc.realm === 'prod') tmpArray[sc.clientId].prod = 1;
+          });
+          //convert back to numeric array
+          for (var items in tmpArray) {
+            output.push(tmpArray[items]);
           }
```
jujaga: Not sure what this loop is really trying to achieve. This looks like you're effectively doing
`output = tmpArray` as you don't use output anywhere else and getting all elements transferred over.

app/frontend/src/components/Stats.vue
```diff
@@ -1,79 +1,113 @@
...
+  mounted() {
+    this.getData();
+  }
 };
 </script>
+
+<style scoped>
+.kc-table .active {
+  width: 10px;
+  height: 10px;
```
jujaga: Use rem/em values whenever possible.

app/src/components/keyCloakServiceClientMgr.js
```diff
@@ -169,17 +169,23 @@ class KeyCloakServiceClientManager {
     }
   }
 
-  async fetchAllClients() {
-    log.info('KeyCloakServiceClientManager.fetchAllClients');
+  async fetchAllClients(kcEnv) {
+
+    log.info('KeyCloakServiceClientManager.fetchAllClients in ' + kcEnv);
```
jujaga: log.debug('KeyCloakServiceClientManager.fetchAllClients',  `Env: ${kcEnv}`);

app/src/components/keyCloakServiceClientMgr.js
```diff
@@ -169,17 +169,23 @@ class KeyCloakServiceClientManager {
...
 
     //get all service clients
-    // NOTE: this seems to return all the service clinets but shows a ERR 400 in the console for all the methods in the reamlAdminSvc class
-    const clients = await this.svc.getClients();
+    let clients = await this.svc.getClients();
```
jujaga: It's not great practice to be reassigning something you get from a response. It's better to either create another const variable and manipulate, or consider function chaining to remove the need altogether.

app/src/components/keyCloakServiceClientMgr.js
```diff
@@ -169,17 +169,23 @@ class KeyCloakServiceClientManager {
...
+
+    // remove those that dont match '*_SERVICE_CLIENT';
+    const regex = '.*_SERVICE_CLIENT$';
+    clients = clients.filter(cl => cl.clientId.match(regex));
```
jujaga: Since we aren't really using this regex anywhere else, you can just combine this into the following line
`clients.filter(cl => cl.clientId.match(/.*_SERVICE_CLIENT$/))`

app/src/routes/v1/keycloak.js
```diff
@@ -0,0 +1,54 @@
...
+    endpoint: realmBaseUrl,
+    username: clientId,
+    password: clientSecret,
+    realm: realmId
+  } = config.get(realmKey);
+
+  const realmSvc = new RealmAdminService({ realmBaseUrl, clientId, clientSecret, realmId });
+  const kcScMgr = new KeyCloakServiceClientManager(realmSvc);
+
+  return kcScMgr.fetchAllClients(kcEnv);
```
loneil: The manager class calls should be agnostic of environment, since it's constructed with the realm service which is set up with the appropriate environment. So it looks like this env is just being passed in to add a static label field. That can be pulled up to this getClientsFromEnv method. 
IE add the env to the returned objects from the fetchAllClients call here rather than sending the env down.

app/src/components/keyCloakServiceClientMgr.js
```diff
@@ -168,6 +168,26 @@ class KeyCloakServiceClientManager {
...
+
+    //get all service clients
+    let clients = await this.svc.getClients();
+
+    // remove those that dont match '*_SERVICE_CLIENT';
+    const regex = '.*_SERVICE_CLIENT$';
+    clients = clients.filter(cl => cl.clientId.match(regex));
+
+    // add the keycloak environment to each service client object in the array
+    clients.forEach(function (cl) {
```
loneil: See comment below about not needing the env passed down.

app/src/components/keyCloakServiceClientMgr.js
```diff
@@ -169,17 +169,23 @@ class KeyCloakServiceClientManager {
...
 
-    // get details for each service client
-    const clientsWithDetails = clients.map(sc => this.makeClientDetails(sc));
+    // add the keycloak environment to each service client object in the array
+    clients.forEach(function (cl) {
+      cl.realm = kcEnv;
+    });
 
-    return await Promise.all(clientsWithDetails);
+    return clients;
```
jujaga: To extend on the above comments, you can choose to do array filter chaining:
```
const clients = await this.svc.getClients();
return clients
    .filter(cl => cl.clientId.match(/.*_SERVICE_CLIENT$/))
    .map(cl => { ...cl, realm: kcEnv });
```
This should be functionally equivalent.

app/frontend/src/components/Clients.vue
```diff
@@ -0,0 +1,113 @@
...
+        dense
+        :headers="headers"
+        :items="serviceClients"
+        :items-per-page="5"
+        :search="search"
+        :loading="loading"
+        loading-text="Loading... Please wait"
+      >
+        <template v-slot:item.dev="{item}">
+          <v-icon v-if="item.dev === 1" color="green">check</v-icon>
```
loneil: Isn't the ennironment a property of the client object itself?
https://getok-dev.pathfinder.gov.bc.ca/pr-149/api/v1/docs#operation/kcServiceClientsGet
So should be able to just check the environment === 'DEV' rather than the setup of adding a flag to each one.

app/src/routes/v1/keycloak.js
```diff
@@ -0,0 +1,54 @@
...
+}
```
jujaga: This function looks out of place in the router middleware level - can't remember where it used to reside however, but it definitely doesn't belong in the router layer.

app/src/routes/v1/keycloak.js
```diff
@@ -0,0 +1,54 @@
...
+keycloakRouter.get('/serviceClients', keycloak.protect('realm:GETOK_ADMIN'), async (req, res) => {
+
+  const [devClients, testClients, prodClients] = await Promise.all([
+    getClientsFromEnv('dev'),
+    getClientsFromEnv('test'),
+    getClientsFromEnv('prod')
+  ]);
+
+  // join them all into one array
+  var result = devClients.concat(testClients, prodClients);
```
jujaga: Since we don't actually care about the discrete difference between dev, test and prod clients, you can do the same thing more succinctly:
```
const result = await Promise.all([
    getClientsFromEnv('dev'),
    getClientsFromEnv('test'),
    getClientsFromEnv('prod')
]).flat();
```
Also - use const or let. Do not use var for variable declarations.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

app/frontend/src/components/Clients.vue
loneil: Actually disregard this above.
However, should use true/false if going this way rather than a 1

app/src/docs/v1.api-spec.yaml
```diff
@@ -169,32 +169,19 @@ paths:
...
-            type: string
       responses:
         "200":
-          description: Returns the Keycloak client details for the acronym specified
+          description: Returns the registered service clients
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/KcClientDetail"
+                $ref: "#/components/schemas/KcClients"
```
loneil: I don't think this is correct is it? Not all the fields from this object are coming back from this new API call right? (since it's a simpler one)
Like "commonServiceRoles", does that get included in this api call? It shows in the docs.

app/frontend/src/components/Clients.vue
```diff
@@ -0,0 +1,113 @@
...
+  watch: {
+    // hide data table progress bar when serviceClients have been returned from backend
+    serviceClients() {
+      this.loading = false;
+    }
+  },
+  methods: {
+    // get table data from frontend service layer
+    getData() {
+      keycloakService.getServiceClients().then(response => {
```
loneil: Can have a catch here to handle an error.

TimCsaky:
these look like awesome suggestions. i'll take a look at them now

app/frontend/src/components/Clients.vue
```diff
@@ -0,0 +1,113 @@
...
+            if (sc.realm === 'dev') tmpArray[sc.clientId].dev = 1;
+            if (sc.realm === 'test') tmpArray[sc.clientId].test = 1;
+            if (sc.realm === 'prod') tmpArray[sc.clientId].prod = 1;
+          });
+          //convert back to numeric array
+          for (var items in tmpArray) {
+            output.push(tmpArray[items]);
+          }
+        }
+        this.serviceClients = output;
```
loneil: Can use a reduce.
`        const res = result.reduce((a, b) => {
      if (!a[b.clientId]) a[b.clientId] = []; //If this type wasn't previously stored
      a[b.clientId].push(b.environment);
      return a;
    }, {});
    const r2 = Object.keys(res).map(k => { return { client: k, environments: res[k] }; });;`
(change names from a, b maybe if you want)

I tried this out in a unit tests and I get 
`    [
      {
        "client": "ZZZZ_SERVICE_CLIENT",
        "environments": [
          "TEST",
          "DEV"
        ]
      },
      {
        "client": "XXX_SERVICE_CLIENT",
        "environments": [
          "DEV",
          "TEST",
          "PROD"
        ]
      }
    ]`

Then you can check if the array contains the ENV in each column in the table to indicate the checkmark

app/frontend/src/components/Clients.vue
loneil: See my comment below about using a reduce/map, then you can check for each of these if the client.environments contains DEV, etc

app/src/components/keyCloakServiceClientMgr.js
loneil: Also this field is called 'environment' in the spec for this new endpoint, not 'realm'

app/src/routes/v1/keycloak.js
TimCsaky: ok. not sure if you can flatten a promise. how about:
```
  const clients = await Promise.all([
    getClientsFromEnv('dev'),
    getClientsFromEnv('test'),
    getClientsFromEnv('prod')
  ]);
  // join them all into one array
  const result = clients.flat();
```

app/src/routes/v1/keycloak.js
TimCsaky: I understand this now. thanks

app/src/routes/v1/keycloak.js
TimCsaky: good call. i think we talked about having it somewhere else. possibly in a services/keycloak.js file. to be discussed no doubt!

app/frontend/src/components/Stats.vue
TimCsaky: good call

app/frontend/src/components/Clients.vue
TimCsaky: hey thanks for taking the time to show this. it does work. i struggled with it for a while. very glad to have this for future reference.

app/src/docs/v1.api-spec.yaml
TimCsaky: I think you're correct. my new api call returns the keycloak api response. 
`{
    id: 'd288f328-3d0e-4bd0-8d89-7bdccface260',
    clientId: 'DGRSC_SERVICE_CLIENT',
    name: 'Doc Gen Showcase',
    description: 'Document Generation Showcase Application',
    surrogateAuthRequired: false,
    enabled: true,
    clientAuthenticatorType: 'client-secret',
    redirectUris: [],
    webOrigins: [],
    notBefore: 0,
    bearerOnly: false,
    consentRequired: false,
    standardFlowEnabled: false,
    implicitFlowEnabled: false,
    directAccessGrantsEnabled: false,
    serviceAccountsEnabled: true,
    authorizationServicesEnabled: true,
    publicClient: false,
    frontchannelLogout: false,
    protocol: 'openid-connect',
    attributes: {
      'saml.assertion.signature': 'false',
      'saml.multivalued.roles': 'false',
      'saml.force.post.binding': 'false',
      'saml.encrypt': 'false',
      'saml.server.signature': 'false',
      'saml.server.signature.keyinfo.ext': 'false',
      'exclude.session.state.from.auth.response': 'false',
      saml_force_name_id_format: 'false',
      'saml.client.signature': 'false',
      'tls.client.certificate.bound.access.tokens': 'false',
      'saml.authnstatement': 'false',
      'display.on.consent.screen': 'false',
      'saml.onetimeuse.condition': 'false'
    },
    authenticationFlowBindingOverrides: {},
    fullScopeAllowed: false,
    nodeReRegistrationTimeout: -1,
    protocolMappers: [ [Object], [Object], [Object] ],
    defaultClientScopes: [ 'web-origins', 'roles' ],
    optionalClientScopes: [],
    access: { view: true, configure: true, manage: true }
  },`

i think it's this one: https://www.keycloak.org/docs-api/5.0/rest-api/index.html#_clientrepresentation 
i could refer to that in our api spec. 
or perhaps map the properties we want from the response to a new object and describe that in a new schema.

TimCsaky: i can fix this on Monday.

app/src/routes/v1/keycloak.js
TimCsaky: I've moved this to a backend 'clients' component

app/src/docs/v1.api-spec.yaml
TimCsaky: i've had another go at the api spec. i'm not sure it's quite right but dont want to spend too long on it if possible

app/frontend/src/components/Clients.vue
TimCsaky: if added an else statement in case no data is returned and done a basic alert on the table. let me know if that's not what you had in mind.

app/src/components/keyCloakServiceClientMgr.js
TimCsaky: I no longer have kcEnv parameter passed to this function.

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-149

app/frontend/src/components/Clients.vue
TimCsaky: catch added

app/src/components/clients.js
```diff
@@ -0,0 +1,40 @@
...
+      endpoint: realmBaseUrl,
+      username: clientId,
+      password: clientSecret,
+      realm: realmId
+    } = config.get(realmKey);
+
+    const realmSvc = new RealmAdminService({ realmBaseUrl, clientId, clientSecret, realmId });
+    const kcScMgr = new KeyCloakServiceClientManager(realmSvc);
+
+    var clients = kcScMgr.fetchAllClients();
```
jujaga: This needs to be a const, not a var

app/src/components/clients.js
```diff
@@ -0,0 +1,40 @@
...
+    clients.then(result => {
+      // add a keycloak environment label to each service client object in the array
+      result.forEach(function (cl) {
+        cl.environment = kcEnv;
+      });
+    });
+
+    return clients;
```
jujaga: The way it is written right now, this function will return clients before the internal forEach is fully executed as clients itself will be an unresolved promise. Since this getClientsFromEnv function is already designated as async, you should instead do something like the following:

```
const clients = await kcScMgr.fetchAllClients();
clients.forEach(cl => cl.environment = kcEnv);

return clients;
```
This will make it so that all clients must have been resolved first before proceeding with the insertion of environment attributes synchronously and returning.

app/src/docs/v1.api-spec.yaml
```diff
@@ -876,16 +876,99 @@ components:
...
+          'saml.server.signature': 'false',
+          'saml.server.signature.keyinfo.ext': 'false',
+          'exclude.session.state.from.auth.response': 'false',
+          'saml_force_name_id_format': 'false',
+          'saml.client.signature': 'false',
+          'tls.client.certificate.bound.access.tokens': 'false',
+          'saml.authnstatement': 'false',
+          'display.on.consent.screen': 'false',
+          'saml.onetimeuse.condition': 'false'
+        }
```
jujaga: While this is technically valid as YAML allows JSON objects, we should try to ensure that this is represented as pure YAML instead. I don't know if it needs to be updated so that saml is its own attribute instead, but I suspect you just don't need the curly braces:

```yaml
example:
  saml.assertion.signature: 'false'
  saml.multivalued.roles: 'false'
  saml.force.post.binding: 'false'
  saml.encrypt: 'false'
  saml.server.signature: 'false'
  saml.server.signature.keyinfo.ext: 'false'
  exclude.session.state.from.auth.response: 'false'
  saml_force_name_id_format: 'false'
  saml.client.signature: 'false'
  tls.client.certificate.bound.access.tokens: 'false'
  saml.authnstatement: 'false'
  display.on.consent.screen: 'false'
  saml.onetimeuse.condition: 'false'
```

app/src/routes/v1/keycloak.js
```diff
@@ -1,24 +1,20 @@
...
-    getClientsFromEnv('test'),
-    getClientsFromEnv('prod')
+  const serviceClients = await Promise.all([
+    clients.getClientsFromEnv('dev'),
+    clients.getClientsFromEnv('test'),
+    clients.getClientsFromEnv('prod')
   ]);
   // join them all into one array
-  const result = clients.flat();
+  const result = serviceClients.flat();
```
jujaga: You don't really need to create another temporary constant to flatten the array when it could be done in-line:
```js
  const result = await Promise.all([
    clients.getClientsFromEnv('dev'),
    clients.getClientsFromEnv('test'),
    clients.getClientsFromEnv('prod')
  ]).flat();
```

app/src/components/clients.js
TimCsaky: thank you! i was trying to figure that one out. i see how that works now

app/src/components/keyCloakServiceClientMgr.js
```diff
@@ -168,6 +168,18 @@ class KeyCloakServiceClientManager {
       return undefined;
     }
   }
+
+  async fetchAllClients() {
+
+    log.info('KeyCloakServiceClientManager.fetchAllClients');
```
jujaga: If this needs to be logged, it needs to be at `debug` level instead of info as all this does is say that this function was invoked at some point in the stack.

app/src/routes/v1/keycloak.js
TimCsaky: yeah i tried that. but console tells me:
`TypeError: Promise.all(...).flat is not a function`

app/src/components/keyCloakServiceClientMgr.js
TimCsaky: that makes sense

app/src/routes/v1/keycloak.js
jujaga: Ah - might be order of operations. Let's leave it as is then.

loneil:
Would consider

- Setting the client name to sort alpha by default in the table
- Expanding the default table size to 10 or 20

But can leave that with Matt's descision

PR #153, opened 2020-03-25
Skeleton structure of API Access components

loneil:
# Description
WIP structure for 
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-766
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-767
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-768
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-769

Add the views and navigation and components for the API access form in skeleton, and the backing Vuex store.
Would like to get feedback/review on this before proceeding with getting all the other stuff in, and to get it integrated into the code base.

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/RequestForm.vue
+++ b/app/frontend/src/components/RequestForm.vue
@@ -88,8 +88,8 @@
       </v-row>
     </v-form>
     <div class="justify-center pb-8">
-      <v-btn class="request-form-cancel-btn BC-Gov-PrimaryButton light float-left" text @click="cancel()">Cancel</v-btn>
-      <v-btn class="request-form-submit-btn BC-Gov-PrimaryButton float-right" text @click="postRegistrationForm()">Submit</v-btn>
+      <v-btn class="BC-Gov-PrimaryButton light float-left" text @click="cancel()">Cancel</v-btn>
+      <v-btn class="BC-Gov-PrimaryButton float-right" text @click="postRegistrationForm()">Submit</v-btn>
     </div>
 
     <BaseDialog v-bind:show="errorOccurred" @close-dialog="errorOccurred = false">
--- a/app/frontend/src/components/apiAccess/Step4.vue
+++ b/app/frontend/src/components/apiAccess/Step4.vue
@@ -6,17 +6,21 @@
     <br />
     <br />
     <div class="text-center">
-      <v-btn class="BC-Gov-PrimaryButton" text @click="finish()">Finish</v-btn>
+      <v-btn class="BC-Gov-PrimaryButton" text @click="finish">Finish</v-btn>
     </div>
   </v-container>
 </template>
 
 <script>
+import { mapMutations } from 'vuex';
+
 export default {
   name: 'ApiAccessStep4',
   methods: {
+    ...mapMutations('apiAccess', ['setStep']),
     finish() {
       this.$router.push({ name: 'MyApps' });
+      this.setStep(1);
     }
   }
 };
--- a/app/frontend/src/components/apiAccess/Step4.vue
+++ b/app/frontend/src/components/apiAccess/Step4.vue
@@ -1,0 +1,23 @@
+<template>
+  <v-container>
+    <h2>API Access for Common Services</h2>Contents to go here
+    <br />
+    <br />
+    <br />
+    <br />
+    <div class="text-center">
+      <v-btn class="BC-Gov-PrimaryButton" text @click="finish()">Finish</v-btn>
+    </div>
+  </v-container>
+</template>
+
+<script>
+export default {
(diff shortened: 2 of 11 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-153

app/frontend/src/views/Application.vue
```diff
@@ -1,12 +1,81 @@
...
+/* Breadcrumb */
+.app-breadcrumb {
+  margin-top: 0.5em;
+  span {
+    padding-right: 1em;
+  }
+}
+
+/* Tab (TODO: if this is used elsewhere in the program, make this style global) */
+.getok-tabs {
```
jujaga: Tabs are currently used in JWTDialog component as well. This may be worth putting into the main .scss.

app/frontend/src/components/RequestForm.vue
```diff
@@ -88,8 +88,8 @@
       </v-row>
     </v-form>
     <div class="justify-center pb-8">
-      <v-btn class="request-form-cancel-btn BC-Gov-PrimaryButton light float-left" text @click="cancel()">Cancel</v-btn>
-      <v-btn class="request-form-submit-btn BC-Gov-PrimaryButton float-right" text @click="postRegistrationForm()">Submit</v-btn>
```
jujaga: These classes were added in to allow the button to be searchable in a future unit test. The intention is to have the unit test be able to "click" on a button, and then ensure that the associated call is invoked.
Granted it isn't exactly needed either at the moment since I can't seem to get `wrapper.find('.request-form-submit-btn').trigger('click')` to actually make a click happen... not sure if it's implementation error or a bug.

app/frontend/src/components/apiAccess/Step4.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+  <v-container>
+    <h2>API Access for Common Services</h2>Contents to go here
+    <br />
+    <br />
+    <br />
+    <br />
+    <div class="text-center">
+      <v-btn class="BC-Gov-PrimaryButton" text @click="finish()">Finish</v-btn>
```
jujaga: IIRC you can do `@click="finish"` without the parentheses and the method will still invoke as it's parameter-less.

app/frontend/src/views/Application.vue
loneil: Do they do vertical or horizontal tabs? The changes so far for this one might have some specificity to vertical tabs with some padding I'll have to put in to match the mockups.
But yeah if we want this general then it should go into the main and then "getok-tabs" can be put on to tabs to opt in to the style (or something like that)
We can circle back comparing the 2 tab usages.

app/frontend/src/components/RequestForm.vue
loneil: Hmmm, ok. I can put them back if we want. Searched and didn't find them referenced anywhere else in the project.
For tests setting an ID might be more appropriate than a class? Not sure.

app/frontend/src/components/apiAccess/Step4.vue
loneil: Oh yeah, you're right.

app/frontend/src/store/modules/apiAccess.js
```diff
@@ -0,0 +1,16 @@
...
+export default {
+  namespaced: true,
+  state: {
+    step: 1
+  },
+  getters: {
+    step: state => state.step
+  },
+  mutations: {
+    setStep: (state, step) => {
```
TimCsaky: i was just reading about passing objects to mutations in the store. presumably you can make a lot of the api calls from in this file too.

app/frontend/src/components/RequestForm.vue
jujaga: As per discussion, let's leave the classes removed for now. If/when we need them for selection, we'll readd them in as id tags for unit tests.

app/frontend/src/views/Application.vue
jujaga: The JWTDialog uses horizontal standard tabs. We can revisit this at a later time to try and standardize.

app/frontend/src/store/modules/apiAccess.js
jujaga: That is correct, though by convention iirc, you'll want to try to do those calls in an action instead of a mutation since a mutation is synchronous and focuses on changing the state, whereas an action can be asynchronous, and is better suited for doing api calls that then call the mutations.

PR #155, opened 2020-03-28
Step 1 to Step 2

loneil:
# Description
Note: to add acronyms to yourself at this point (we will seed data later in PRs), go the admin page and in the API tester run
`acronyms/<acr>/addUser/<you@idir>`
like
`acronyms/MSSC/addUser/loneil@idir`
The acronym does not have to exist in the system, this will add it if it's not there.

Step 1 to Step 2 of the API access flow, doing this PR now to have a smaller one for the remainder of of the API access pages (23 file changes already).

Should load step 1 with service client access logic and go through step 2 up until the submit point, where it won't execute yet.

## Types of changes

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-155

app/frontend/src/components/apiAccess/Step2.vue
```diff
@@ -79,20 +81,50 @@ export default {
...
   computed: {
-    ...mapGetters('apiAccess', [
-      'acronym',
-      'appName',
-      'appDescription',
-      'environment'
-    ])
+    ...mapGetters('apiAccess', ['acronym', 'environment']),
+    // TODO: 2-way binding to vuex state with v-model (https://vuex.vuejs.org/guide/forms.html), but there's got to be a less verbose way...
+    // Going with this to get features working for now
```
jujaga: This is actually one of the potential patterns we can stick with, except in this case we'd no longer use ...mapGetters(...) and just make sure our computed properties directly get and commit from the store.
Ref: https://ypereirareis.github.io/blog/2017/04/25/vuejs-two-way-data-binding-state-management-vuex-strict-mode/#edit-two-way-computed-property

app/frontend/src/components/apiAccess/ApiAccess.vue
```diff
@@ -45,11 +52,24 @@ export default {
...
   computed: {
     ...mapGetters('apiAccess', ['step'])
   },
   methods: {
-    ...mapMutations('apiAccess', ['setStep'])
+    ...mapMutations('apiAccess', ['setStep', 'setAcronym']),
+    ...mapActions('apiAccess', ['setStep', 'getAcronymClientStatus'])
+  },
+  created() {
+    this.$store.registerModule('apiAccess', apiAccess);
```
jujaga: I'm going to suggest we register and unregister the apiAccess state module only in the view layer. This makes it so that we know that when we are looking at a specific route view, we know the Vuex state is mounted and available for any children components.

As an extension, for every registerModule, there must also be an equivalent unregisterModule in the beforeDestroy lifecycle. Reference the `MyApps.vue` view for an example. In this case, I believe this registerModule will need to reside under the `Application.vue` view, and you'll need to add the equivalent unregisterModule on the beforeDestroy lifecycle for it.

app/frontend/src/components/apiAccess/Step1.vue
```diff
@@ -1,21 +1,89 @@
...
-    <br />
-    <br />
-    <br />
-    <v-btn color="primary" @click="setStep(2)">Get Token for Dev</v-btn>
+  <v-container class="pl-0">
+    <h2 class="pb-8">API Access for Common Services</h2>
+    <p>Your application has a service client with access to:</p>
+    <ul>
+      <li>Common Hosted Email Service</li>
+      <li>Common Document Generation Service</li>
```
jujaga: May be worth alphabetizing this list, but that's a UX decision.

app/frontend/src/components/base/BaseDialog.vue
```diff
@@ -23,9 +23,19 @@
         </v-card-text>
       </div>
       <v-card-actions class="justify-center">
-        <v-btn class="BC-Gov-PrimaryButton mb-5" text @click="closeDialog">
-          <slot name="button-text">OK</slot>
-        </v-btn>
+        <div v-if="type == 'OK'">
```
jujaga: ===

app/frontend/src/components/base/BaseDialog.vue
```diff
@@ -23,9 +23,19 @@
...
+          </v-btn>
+        </div>
+        <div v-else-if="type === 'CONTINUE'">
+          <v-btn class="BC-Gov-PrimaryButton light mb-5 mr-5" text @click="closeDialog">
+            <slot name="button-text">Cancel</slot>
+          </v-btn>
+          <v-btn class="BC-Gov-PrimaryButton mb-5" text @click="continueDialog">
+            <slot name="button-text">Continue</slot>
+          </v-btn>
+        </div>
```
jujaga: I'm wondering if we should be making the v-card-actions be more flexible instead and have this entire section be the slot instead of just the button and text. I ran into a similar issue when I was using BaseDialog for the admin panel components. We can allow for more flexibility, but at the potential cost of needing to custom style on the parent level - something we should consider and chat about.

app/frontend/src/services/acronymService.js
```diff
@@ -0,0 +1,18 @@
+import { getokAxios } from '@/services/interceptors';
+
+import { ApiRoutes } from '@/utils/constants';
+
+export default {
+  /**
+   * @function getServiceClients
+   * Fetch the service clients for the acronym
+   * @param {string} acronym app acronym
+   */
```
jujaga: `@returns {Promise} An axios response`

app/frontend/src/utils/constants.js
```diff
@@ -20,6 +20,17 @@ export const FieldValidations = Object.freeze({
...
+export const KcEnv = Object.freeze({
+  DEV: 'DEV',
+  TEST: 'TEST',
+  PROD: 'PROD'
+});
+
+export const KcClientStatus = Object.freeze({
+  AVAILABLE: 'Available',
+  NOT: 'Not Available'
+});
```
jujaga: Are we intending on usins these constants elsewhere in the codebase? I saw you refactored this out of the component. Mixed feelings on whether the extraction is needed or not.

openshift/commonPipeline.groovy
```diff
@@ -123,7 +123,7 @@ def runStageDeploy(String stageEnv, String projectEnv, String hostEnv, String pa
       openshift.selector('cm', "getok-server-config").exists() &&
       openshift.selector('secret', "getok-keycloak-secret").exists() &&
       openshift.selector('secret', "getok-sc-ches-secret").exists() &&
-      openshift.selector('secret', "getok-sc-keycloakint-secret").exists() &&
+      openshift.selector('secret', "getok-sc-keycloakdev-secret").exists() &&
```
jujaga: I think we should have this be `getok-sc-keycloak-dev-secret` instead (and updated for everything else referencing it in order to be consistent with all the other secret nomenclatures.

openshift/app.dc.yaml
```diff
@@ -162,6 +162,36 @@ objects:
                     secretKeyRef:
                       key: password
                       name: getok-sc-ches-secret
+                - name: SC_KC_DEV_USERNAME
+                  valueFrom:
+                    secretKeyRef:
+                      key: username
+                      name: getok-sc-keycloakdev-secret
```
jujaga: I think we should have this be `getok-sc-keycloak-dev-secret` instead (and updated for everything else referencing it in order to be consistent with all the other secret nomenclatures.

openshift/README.md
```diff
@@ -74,8 +71,7 @@ oc create -n k8vopl-<env> secret generic getok-sc-ches-secret \
 ```
 
 ```sh
-oc create -n k8vopl-<env> secret generic getok-sc-keycloakint-secret \
-  --type=kubernetes.io/basic-auth \
+oc create -n k8vopl-<env> secret generic getok-sc-keycloakdev-secret type=kubernetes.io/basic-auth \
```
jujaga: I think we should have this be `getok-sc-keycloak-dev-secret` instead (and updated for everything else referencing it in order to be consistent with all the other secret nomenclatures.

app/src/routes/v1/acronyms.js
```diff
@@ -0,0 +1,60 @@
...
+    if (!req.params.appAcronym || !req.params.username) {
+      res.status(400).json({
+        message: 'Must supply app acronym and user (ex: myname@idir)'
+      });
+      return res;
+    }
+
+    try {
+      const token = req.headers.authorization.split(' ')[1];
+      const response = await acronyms.registerUserToAcronym(token, req.kauth.grant.access_token.content.iss, req.params.appAcronym, req.params.username);
```
jujaga: IIRC we already have the jsonwebtoken library as part of the codebase so we should try to leverage that instead if possible - refer to the checks.js component and see if it's applicable (I think I was using the jwt.decode(...) functionality).

app/src/components/utils.js
```diff
@@ -66,6 +69,25 @@ const utils = {
...
+    const {
+      endpoint: realmBaseUrl,
+      username: clientId,
+      password: clientSecret,
+      realm: realmId
+    } = config.get(realmKey);
+    const realmSvc = new RealmAdminService({ realmBaseUrl, clientId, clientSecret, realmId });
+    const kcScMgr = new KeyCloakServiceClientManager(realmSvc);
+    return kcScMgr.fetchClients(acronyms);
+  }
```
jujaga: Let's have this be declared above getKeyCloakToken (alphabetical ordering)

app/frontend/src/utils/util.js
```diff
@@ -0,0 +1,15 @@
+
```
jujaga: This file should be utils.js (plural)
Unneeded whitespace line here.

app/frontend/src/utils/util.js
```diff
@@ -0,0 +1,15 @@
...
+ * Returns the appropriate DOM for client Available or Not available (wording and CSS)
+ * @param {string} envLabel Which environment (DEV, TEST, PROD)
+ * @param {boolean} status Whether the service client is there
+ * @returns {string} HTML
+ */
+export function buildClientStatusSpan(envLabel, status) {
+  const cls = status ? 'green--text' : '';
+  const txt = status ? KcClientStatus.AVAILABLE : KcClientStatus.NOT;
+  return `${envLabel}: <span class="${cls}">${txt}</span>`;
+}
```
jujaga: Not quite to convinced this should be abstracted out as a common utils function - most of the logic is basically checking if the text is green, and if it is, saying it's also available. This may be business logic better suited to live in the component it came from, unless this is being utilized in multiple components.

app/frontend/src/components/apiAccess/ApiAccess.vue
loneil: Ok I'll try that out

app/frontend/src/components/apiAccess/Step1.vue
loneil: Noting for discussion with Mina later

app/frontend/src/components/apiAccess/Step2.vue
loneil: Yeah I've seen it for form access/2 way binding. I'd keep the getters for just read since it's more succinct.
So it's probably the right way as you say, I still don't like the overkill syntax :)
I think this is one of the main complaints about Vuex and I've seen it mentioned there could be some big changes in future version.

app/frontend/src/components/base/BaseDialog.vue
loneil: Fixed

app/frontend/src/components/base/BaseDialog.vue
loneil: Yeah possibly, want to see where we can improve this now that we're using it for more functionality.
Can discuss coming next sprint maybe.

app/frontend/src/services/acronymService.js
loneil: fixed

app/frontend/src/utils/constants.js
loneil: Yeah I'd refactored it out and was using it as a comparison, so it was appropriate. But then I changed implementation.
Let's leave it for now, and once I wrap all the other code for this I'll see if it's used. 
Added a TODO comment

app/frontend/src/utils/util.js
loneil: Yeah it's used in 2 different components so I refactored it out. I'm not crazy about View building in a utils class...
... oh wait... it should just be its own small component then with props passed in.
If you think that makes sense let me know. I'll do it in my next PR.

app/frontend/src/utils/util.js
loneil: Might just remove this file. See comment below (if so will leave until next PR)

app/src/components/utils.js
loneil: Fixed

openshift/README.md
loneil: Can we do this with next PR, might not want to clash with other PRs using the configs.

app/frontend/src/components/apiAccess/Step2.vue
jujaga: Yea - feels a bit overkill for syntax I agree, but at least it achieves what we need it to do.

app/frontend/src/components/apiAccess/ApiAccess.vue
loneil: Works nicely. Thanks!

openshift/README.md
jujaga: Ok. I can wrap this as a part of the upcoming db seed changes as that will do some pipeline modifications anyways.

app/frontend/src/utils/util.js
jujaga: This would be one of those fantastic cases where we can use a "functional SFC" and pass in via prop. I think that may make more sense - if nothing else that should be something we can investigate. Here's a pretty good article on it and it has an example SFC that this could match with:
https://dev.to/fdietz/vue-js-functional-components-2495

app/frontend/src/utils/util.js
jujaga: This would be one of those fantastic cases where we can use a "functional SFC" and pass in via prop. I think that may make more sense - if nothing else that should be something we can investigate. Here's a pretty good article on it and it has an example SFC that this could match with:
https://dev.to/fdietz/vue-js-functional-components-2495

openshift/app.dc.yaml
jujaga: Ok. I can wrap this as a part of the upcoming db seed changes as that will do some pipeline modifications anyways.

openshift/commonPipeline.groovy
jujaga: Ok. I can wrap this as a part of the upcoming db seed changes as that will do some pipeline modifications anyways.

TimCsaky:
You've done a great job with this. As usual i'm taking a while to read through the code and figure out what's happening which is very useful for me. I wish i had more constructive input.

app/frontend/src/utils/util.js
loneil: Will read up on that for next PR.

app/src/routes/v1/acronyms.js
loneil: The kauth structure is added to the request by Keycloak Connect so we can access the token claims from that without having to do anything with the raw token

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-155

PR #159, opened 2020-04-01
API access step4

loneil:
# Description

## Types of changes
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-769

UI and styling for API access step 4

## Checklist

### DIFF
--- a/app/frontend/src/assets/scss/style.scss
+++ b/app/frontend/src/assets/scss/style.scss
@@ -17,6 +17,17 @@ p {
   line-height: 1.6;
 }
 
+a {
+  text-decoration: none;
+  &:hover {
+    text-decoration: underline;
+    button {
+      display: inline-block;
+      text-decoration: none;
+    }
+  }
+}
+
 
 /* sticky footer */
 html,
@@ -59,4 +70,5 @@ body > .v-content {
 
 .BC-Gov-IconButton.v-btn {
   color: #003366 !important;
+  text-decoration: none;
 }
--- a/app/frontend/src/components/apiAccess/Step4.vue
+++ b/app/frontend/src/components/apiAccess/Step4.vue
@@ -1,21 +1,70 @@
 <template>
   <v-container>
-    <h2>API Access for Common Services</h2>Contents to go here
-    <br />
-    <br />
-    <br />
-    <br />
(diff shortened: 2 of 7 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-159

app/frontend/src/assets/scss/style.scss
```diff
@@ -17,6 +17,17 @@ p {
...
 
+a {
+  text-decoration: none;
+  &:hover {
+    text-decoration: underline;
+    button {
+      display: inline-block;
+      text-decoration: none;
+    }
+  }
```
jujaga, line 28: This doesn't appear to be underlining only the text portion of the button at all on either Firefox or Chrome - unless that is intended?

app/frontend/src/components/apiAccess/Step4.vue
```diff
@@ -1,21 +1,70 @@
...
+import {
+  CommonServiceList,
+  CommonServiceTypes
+} from '@/utils/commonServices.js';
+
 export default {
   name: 'ApiAccessStep4',
+  data() {
+    return {
+      KeycloakCommonServiceList: CommonServiceList.filter(
```
jujaga, line 63: I'm half tempted to say this should be a computed instead of data as we don't directly manipulate this piece of data at all.

loneil:
@parc-jason good catch, they work on local but not on the new PR path structure. Will make a ticket to look at it (should be fine in master)

app/frontend/src/assets/scss/style.scss
loneil: This is intended, it's not supposed to underline in buttons (will confirm with Mina when I discuss some other UI stuff and the styling ticket all underline behaviour)

app/frontend/src/components/apiAccess/Step4.vue
loneil: Doesn't work in a computed. Getting the import in as something referrable in computed and methods/etc requires putting it in the data block anywyas, so it's one-lined there.

PR #160, opened 2020-04-02
Acronyms/users endpoint (spec approval)

loneil:
# Description
Getting a quick feedback on the spec for an endpoint that will serve to populate the table on this page
https://www.figma.com/proto/PqqxnXfahmDGJYC74fj1bQ/onboarding_v5?node-id=30%3A926&scaling=min-zoom

So for an acronym, get me all the users associated and their relevant details.
Will call DB for UserAcronym mapping, then KC for all the user fields.

Docs:
https://getok-dev.pathfinder.gov.bc.ca/pr-160/api/v1/docs#operation/getAcronymUsers

## Types of changes

## Checklist

### DIFF
--- a/app/src/docs/v1.api-spec.yaml
+++ b/app/src/docs/v1.api-spec.yaml
@@ -117,6 +117,38 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  "/acronyms/{acronym}/users":
+    get:
+      summary: Returns all Users from the GETOK database associated with the acronym with their user data from Keycloak
+      operationId: getAcronymUsers
+      tags:
+        - Acronyms
+      parameters:
+        - name: acronym
+          in: path
+          description: Name of the project
+          required: true
+          schema:
+            type: string
+      responses:
+        "200":
+          description: User list
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AcronymUsersList"
+        "401":
+          $ref: "#/components/responses/UnauthorizedError"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "404":
+          $ref: "#/components/responses/NotFound"
+        default:
+          description: Unexpected error
+          content:
+            application/json:
+              schema:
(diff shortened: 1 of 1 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-160

app/src/docs/v1.api-spec.yaml
```diff
@@ -588,6 +620,17 @@ components:
         permissionWebadeNrosDms:
           type: boolean
           example: false
+    AcronymUsersList:
+      type: array
+      items:
+        $ref: "#/components/schemas/AcronymUserMapping"
+    AcronymUserMapping:
+      type: object
+      properties:
```
jujaga, line 629: I think we should add an `"acronym": "MSSC"` property because when I look at the other two /acronyms/* endpoints they both have the input parameter also being emitted in the return JSON object - it's more mainly for consistency with the other existing endpoints, and it doesn't hurt to return back to the caller an extra confirmation that the requested object is what they actually wanted.

app/src/docs/v1.api-spec.yaml
```diff
@@ -1085,6 +1128,37 @@ components:
...
+          description: First name from KC
+          example: John
+        lastName:
+          type: string
+          description: Last name from KC
+          example: Smith
+        email:
+          type: string
+          description: Email from KC
+          example: jsmith@gov.bc.ca
```
jujaga, line 1161: Ideally should be alphabetical unless there is a logical reason on ordering the attributes this way.

loneil:
> Looks good, nice and simple.
> For the immediate future, we have no concept of roles or permissions within the team (ie admin, promoter, viewer) - everyone is equal? And all users are IDIR?

@parc-jason Yeah exactly, we have an owner field in the DB but unused until future features. Users don't necessarily have to be IDIRs, but that's all we're supporting for now.

PR #164, opened 2020-04-06
Team tab table

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-774
Populate table in the Team tab on load of components with the result from the acronyms/<acr>/users call.

Update vuetify.

To test, go to one of the pre-populated acronyms in the PR and look in the Team tab.
Can see a fresh acronym by adding new one at the admin API tester with
/acronyms/<new acr>/yourname@idir

## Types of changes

## Checklist

### DIFF
--- a/app/frontend/src/components/Team.vue
+++ b/app/frontend/src/components/Team.vue
@@ -3,7 +3,7 @@
     <p class="text-right">
       <strong>
         <v-icon class="pr-2">group</v-icon>Team members
-        <span v-if="usersLoaded">({{users.length}})</span>
+        <span v-if="usersLoaded">({{ users.length }})</span>
         <v-tooltip bottom open-delay="1000">
           <template v-slot:activator="{ on }">
             <v-btn v-on="on" icon color="primary" @click="addDialog = true">
@@ -48,13 +48,13 @@
       </v-container>
     </v-skeleton-loader>
 
-    <BaseDialog :show="addDialog" width="500" @close-dialog="addDialog = false">
+    <BaseDialog :show="addDialog" @close-dialog="addDialog = false">
       <template v-slot:icon>
-        <v-icon large color="blue darken-2">info</v-icon>
+        <v-icon large color="primary">info</v-icon>
       </template>
       <template v-slot:text>
         <h4 class="mb-5">Invitation functionality coming soon</h4>
-        <p>To add a new user for your application, please instruct them to log and use the Request Account page to request access to your application.</p>
+        <p>To add a new user for your application, please instruct them to log in and use the Request Account page to request access to your application.</p>
         <p>The Common Service Showcase team will action the request to add them once submitted.</p>
       </template>
     </BaseDialog>
@@ -79,7 +79,7 @@ export default {
     try {
       const res = await acronymService.getUsers(this.acronym);
       this.users = res.data.map(u => {
-        return{
+        return {
           name: `${u.user.firstName} ${u.user.lastName}`,
           idir: u.user.username,
           email: u.user.email
--- a/app/frontend/src/components/Team.vue
+++ b/app/frontend/src/components/Team.vue
@@ -1,0 +1,122 @@
+<template>
+  <v-container>
+    <p class="text-right">
+      <strong>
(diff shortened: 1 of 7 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-164

app/frontend/src/components/Team.vue
```diff
@@ -0,0 +1,108 @@
+<template>
+  <v-container>
+    <p class="text-right">
+      <strong>
+        <v-icon class="pr-2">group</v-icon>Team members
+        <span v-if="usersLoaded">({{users.length}})</span>
```
jujaga, line 6: Style: `{{ users.length }}` Should have spaces between the brackets and contents to make it easier to catch visually.

app/frontend/src/components/Team.vue
```diff
@@ -0,0 +1,108 @@
...
+        <v-tooltip bottom open-delay="1000">
```
jujaga, line 7: There a reason why we're doing a delay of 1 second as opposed to immediate?
Consideration - would the tooltip look better if it was top instead of bottom?

app/frontend/src/components/Team.vue
```diff
@@ -0,0 +1,108 @@
...
+                <td>{{ item.idir }}</td>
+                <td>{{ item.email }}</td>
+              </tr>
+            </tbody>
+          </template>
+        </v-simple-table>
+      </v-container>
+    </v-skeleton-loader>
+
+    <BaseDialog :show="addDialog" width="500" @close-dialog="addDialog = false">
```
jujaga, line 51: Width doesn't need to be specified here as it's already a prop default

app/frontend/src/components/Team.vue
```diff
@@ -0,0 +1,108 @@
...
+      <template v-slot:icon>
+        <v-icon large color="blue darken-2">info</v-icon>
```
jujaga, line 53: Any reason why we're using this color instead of just "primary"?

app/frontend/src/components/Team.vue
```diff
@@ -0,0 +1,108 @@
...
+      errorLoadingUsers: false,
+      users: [],
+      usersLoaded: false
+    };
+  },
+  async mounted() {
+    try {
+      const res = await acronymService.getUsers(this.acronym);
+      this.users = res.data.map(u => {
+        return{
```
jujaga, line 82: Space beteen n and {

app/frontend/tests/unit/services/acronymService.spec.js
```diff
@@ -0,0 +1,75 @@
...
+    acronymId: '123-456',
+    acronym: 'MSSC',
+    name: '',
+    description: 'None Entered',
+    permissionWebade: false,
+    permissionWebadeNrosDms: false,
+    createdAt: '2019-08-15T20:14:41.966Z',
+    updatedAt: '2019-08-15T20:14:41.966Z',
+    deletedAt: null
+  };
```
jujaga, line 28: As a note, we probably don't need this to be "complete" data instance as we don't actually test for pieces of data in there - only that this specific object is the body return. No action required.

app/frontend/src/components/Team.vue
loneil: Fixed

app/frontend/src/components/Team.vue
loneil: Immediate looked annoying with it popping up as soon as you go to click.
Will consult with Mina about these when we do more UI checking later.

app/frontend/src/components/Team.vue
loneil: Fixed

app/frontend/src/components/Team.vue
loneil: Should be primary, fixed

app/frontend/src/components/Team.vue
loneil: Fixed

PR #167, opened 2020-04-07
Client status/buttons wording changes

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-926

Change to a 3-state client status (see ticket). Alterations to My Applications, and API Access step 1.

Refactor status to component instead of the util function.
Wanted to do a functional single file component (https://itnext.io/whats-the-deal-with-functional-components-in-vue-js-513a31eb72b0), but requires computed field calculations based on passed in props so not appropriate for this use.

Update vuetify

## Types of changes

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-167

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-167

app/frontend/src/components/ApplicationList.vue
```diff
@@ -45,23 +49,26 @@
...
+import ClientStatus from '@/components/apiAccess/ClientStatus.vue';
 
 export default {
   name: 'ApplicationList',
+  components: {
+    ClientStatus
+  },
+  data() {
+    return {
+      envs: KcEnv
```
jujaga, line 61: This should be a computed - especially if it's supposed to be a constant resource.

app/frontend/src/components/apiAccess/ClientStatus.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <div>
+    {{ env }}:
```
jujaga, line 3: Should this be in its own span object?

app/frontend/src/components/apiAccess/ClientStatus.vue
```diff
@@ -0,0 +1,54 @@
...
+    <span :class="display.class">{{display.text}}</span>
```
jujaga, line 4: Space between the curly braces and content

app/frontend/src/components/apiAccess/ClientStatus.vue
```diff
@@ -0,0 +1,54 @@
...
+          } else {
+            label = this.clientStatuses.dev ? KcClientStatus.READY : KcClientStatus.NOT;
+          }
+        } else if(this.env === KcEnv.PROD) {
+          if(this.clientStatuses.prod) {
+            label = KcClientStatus.CREATED;
+          } else {
+            label = this.clientStatuses.test ? KcClientStatus.READY : KcClientStatus.NOT;
+          }
+        }
```
jujaga, line 44: If and ( should have a space between.

Half tempted to suggest that this should be a switch instead, considering there's a few repeated else statements. Another option is to just have an auxiliary function which takes in the environment and status as parameters, and outputs the expected constant label instead.

loneil:
Feel free to delete clients for WORG and TEMP (careful only mess with those) in YAMS realms to test out statuses

app/frontend/src/components/apiAccess/Step1.vue
```diff
@@ -70,23 +79,25 @@
...
 export default {
   name: 'ApiAccessStep1',
+  components: {
+    ClientStatus
+  },
   data() {
     return {
       KcClientStatus: KcClientStatus,
-      env: KcEnv
+      envs: KcEnv
```
jujaga, line 93: These should probably be computed as they are meant to be constants.

app/frontend/src/components/ApplicationList.vue
loneil: I guess

app/frontend/src/components/apiAccess/ClientStatus.vue
loneil: No doesn't need to be inline or block level, just text

app/frontend/src/components/apiAccess/ClientStatus.vue
loneil: Fixed

app/frontend/src/components/apiAccess/ClientStatus.vue
loneil: Fixed spacing, would do switch if it was 1 more else if. Leaving as ifs

app/frontend/src/components/apiAccess/Step1.vue
loneil: Fixed

PR #169, opened 2020-04-07
Admin stats

TimCsaky:
# Description

reference: [ticket 856](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-856)
I've finally added a few extra details to the admin 'service client table'.
details to show:
- the dates that app was promoted to each environment
- users in the team for each app
- full name for the application

note: This is just a first draft. I will be adding catches, fixing my bad syntax, adding some unit tests etc tomorrow. I'm more keen to make sure that the data is coming from the right place.
The dates of promotion are from the lifecycle table which I know we will plan to change later.

I did want to get all teh data in one go and re-purpose it for other metrics (passed down to child components) but that hasnt happenned yet.

## Types of changes

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works


app/frontend/src/components/admin/Clients.vue
```diff
@@ -76,19 +122,17 @@ export default {
...
-          const clients = Object.keys(reduced).map(k => {
+          const data = response.data;
+          const clients = Object.keys(data).map(k => {
+            let client = data[k];
             return {
-              name: k,
-              dev: reduced[k].includes('dev'),
-              test: reduced[k].includes('test'),
-              prod: reduced[k].includes('prod')
+              acronym: client.acronym,
```
TimCsaky, line 129: i've changeed the structure of the data from the backend so this mapping to data table is different

app/frontend/src/components/admin/Clients.vue
```diff
@@ -12,20 +14,62 @@
...
+            <table class="kc-nested-table">
+              <!-- environments -->
+              <tr>
+                <td width="40px"></td>
+                <td>
+                  <strong>created:</strong>
+                </td>
+                <td>{{ (item.environments.DEV) ? new Date(item.environments.DEV.created).toLocaleDateString() : '' }}</td>
+                <td>{{ (item.environments.TEST) ? new Date(item.environments.TEST.created).toLocaleDateString() : '' }}</td>
+                <td>{{ (item.environments.PROD) ? new Date(item.environments.PROD.created).toLocaleDateString() : '' }}</td>
```
TimCsaky, line 45: I know this is not pretty

app/frontend/src/components/admin/Clients.vue
```diff
@@ -106,10 +150,44 @@ export default {
...
+}
+.kc-table >>> tr.v-data-table__expanded.v-data-table__expanded__content {
+  -webkit-box-shadow: none !important;
+  box-shadow: none !important;
+}
+.kc-table tr.last td {
+  border-bottom: 1px !important;
+  padding-bottom: 10px;
+}
+.kc-nested-table {
```
TimCsaky, line 184: i have some layout issues complicated by vuetify.
i may need to re-think the layout of the expanded row. will check with Mina

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-169

loneil:
Dates aren't rendering and alignment is borked
image

app/frontend/src/components/admin/Clients.vue
```diff
@@ -12,20 +14,62 @@
...
+        :expanded.sync="expanded"
+        item-key="acronym"
+        show-expand
       >
+        <template v-slot:expanded-item="{ headers, item }">
+          <td :colspan="headers.length">
+            <table class="kc-nested-table">
+              <!-- environments -->
+              <tr>
+                <td width="40px"></td>
```
loneil, line 39: I'm not sure using a table inside this table row is going to work out (I might be wrong...). Especially with set pixel widths, as the column sizes for the parent table will vary with screen size. See my attached screenshot in a comment on this pr.

TimCsaky: yeah. i'll do something with divs

TimCsaky:
> 
> 
> Dates aren't rendering and alignment is borked
> image

i believe only the WORG and TEMP clients will have data in the database. I dont know what's up with the invalid date yet

app/src/routes/v1/keycloak.js
```diff
@@ -8,28 +8,88 @@ const Problem = require('api-problem');
...
+      const acronymnDetailsFromDb = await acronyms.getAcronym(obj.acronym);
+
+      // if a corresponding acronym in db (there wont be in local dev enviroment)
+      if (acronymnDetailsFromDb) {
+
+        obj.acronymnDetails = acronymnDetailsFromDb.dataValues;
+
+        // promotions (from lifecycle table for now)
+        const promotions = await lifecycleService.findLatestPromotions(obj.acronymnDetails.acronymId);
+        promotions.forEach(function (k) {
```
jujaga, line 71: This can be an arrow function as there isn't any context sensitive binds.
```
promotions.forEach(function (k) {
```

app/src/routes/v1/keycloak.js
```diff
@@ -8,28 +8,88 @@ const Problem = require('api-problem');
...
+          obj.environments[k.dataValues.env].created = k.dataValues.createdAt;
+          obj.environments[k.dataValues.env].updated = k.dataValues.updatedAt;
+        });
+
+        // users
+        const usersList = await acronyms.getUsers(obj.acronym);
+        obj.users = usersList;
```
jujaga, line 78: This can be collapsed into a one line await assignment.
```
obj users = await acronyms.getUsers(obj.acronym);
```

loneil:
Users will have to be vertical or some other layout as well
image

app/src/routes/v1/keycloak.js
```diff
@@ -8,28 +8,88 @@ const Problem = require('api-problem');
...
+
+  getDetails(reduced)
+    .then(data => {
+      if (data === null) {
+        return new Problem(404).send(res);
+      } else {
+        res.status(200).json(data);
+      }
+    });
 });
```
jujaga, line 93: There is a lot of new business logic added here in the router layer. This logic should really be in its own component as we want the routing layer to focus on figuring out what needs to be achieved to satisfy an incoming request, whereas a component focuses on what logic needs to be achieved in order to provide the callee with the information they need.

jujaga:
There's a bit of formatting strangeness happening on the users list/array.
table formatting

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-169

app/frontend/src/components/admin/ServiceClients.vue
```diff
@@ -0,0 +1,241 @@
...
+        loading-text="Loading... Please wait"
+        :expanded.sync="expanded"
+        item-key="acronym"
+        show-expand
+
+      >
+        <!-- environments -->
+        <template v-slot:item.dev="{item}">
+          <div v-if="item.dev">
+            <v-tooltip top>
```
TimCsaky, line 37: promotion date shows in tooltip. would be nice to have this tool tip appear on click event as well for mobile users but i dont think that's working

TimCsaky:
obviously only the clients stored in the pr's database will show the extended details (TEMP and WORK i think)

app/frontend/src/components/admin/ServiceClients.vue
jujaga: You can probably add some toggle logic by binding visibility to a v-model. Caveat is when doing this, I suspect the default hover event will no longer be listened to, so you'd have to re-implement that behavior in addition to the on-click toggle.
Reference: https://vuetifyjs.com/en/components/tooltips/#visibility

app/frontend/src/components/admin/ServiceClients.vue
```diff
@@ -0,0 +1,231 @@
...
+            </div>
+
+            <div class="kc-expanded">
+              <!-- users -->
+              <strong>Users: </strong>
+              <ul>
+                <li
+                  v-for="(user, index) in item.users"
+                  :key="index"
+                >{{ user.user.firstName + ' ' + user.user.lastName + ((item.users.length > (index + 1)) ? ',' : '') }}</li>
```
jujaga, line 88: You probably don't need that many parentheses for evaluating where in the array it is - this expression would be equivalent and suffice:
`(index <= item.users.length) ? ',' : ''`

app/frontend/src/components/admin/ServiceClients.vue
```diff
@@ -0,0 +1,231 @@
...
+              </template>
+              <span>Updated: {{ formatEnvDate(item.environments.PROD) }}</span>
+            </v-tooltip>
+          </div>
+        </template>
+
+        <!-- expanded row -->
+        <template v-slot:expanded-item="{ headers, item }">
+          <td :class="[responsiveCell]"></td>
+          <td :colspan="headers.length - 1" :class="[responsiveCell]">
```
jujaga, line 68: You'd have to double check, but seeing as you want the responsiveCell behavior on all table cells, you may be able to just pull that class out into its own div as a parent of it all.
```
<div :class="[responsiveCell]">
  <td ...
  ...
</div>
```

app/frontend/src/components/admin/ServiceClients.vue
```diff
@@ -0,0 +1,231 @@
...
+    };
+  },
+  methods: {
+
+    formatEnvDate(env) {
+      return (env && env.created) ?  this.formatDate(env.created): '-';
+    },
+    formatDate(date){
+      return new Date(date).toLocaleString();
+    },
```
jujaga, line 134: These two functions can probably be chained together into one - this is likely why the tooltip and created date outputs are inconsistent as you're calling them both differently.

app/frontend/src/components/admin/ServiceClients.vue
```diff
@@ -0,0 +1,231 @@
...
+              acronymDetails: client.acronymDetails,
+              dev: client.environments.DEV,
+              test: client.environments.TEST,
+              prod: client.environments.PROD,
+              environments: client.environments,
+              users: client.users
+            };
+          });
+
+          if (clients.length == 0) {
```
jujaga, line 155:
```
if (!clients.length) {
```

app/frontend/src/components/admin/ServiceClients.vue
```diff
@@ -0,0 +1,231 @@
...
+            this.showTableAlert('info', 'No Service Clients found');
+          }
+          this.serviceClients = clients;
+        })
+        .catch(() => {
+          this.showTableAlert('error', 'No response from server');
```
jujaga, line 161: Please add a //TODO note regarding global error notification behavior as this will be relevant if/when we tackle that refactor.

app/frontend/src/components/admin/ServiceClients.vue
TimCsaky: Your code wouldnt work for showing commas.
this would work:
`( (index + 1)  < item.users.length ) ? ',' : ''
`
but its basically what i have

app/frontend/src/components/admin/ServiceClients.vue
TimCsaky: not sure i understand. the class '.v-data-table__mobile-table-row' needs to be applied to all <td> elements on small screens. thats how the vuetify datatable columns are responsive

app/frontend/src/components/admin/ServiceClients.vue
jujaga: Let's just run with that then. Either way, generally it's better to do conditionals with the first term being the value to check, and the second term being the constant you're comparing against (which is what your latest example has. I'd still remove the inner-most parentheses though:
```
( index + 1 < item.users.length ) ? ',' : ''
```

app/frontend/src/components/admin/ServiceClients.vue
```diff
@@ -127,10 +127,11 @@ export default {
   methods: {
 
     formatEnvDate(env) {
-      return (env && env.created) ?  this.formatDate(env.created): '-';
+      return (env && env.created) ?  this.formatDate(env.created): 'N/A';
     },
     formatDate(date){
-      return new Date(date).toLocaleString();
+
```
jujaga, line 133: Unnecessary whitespace line

bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-169

app/src/components/clients.js
```diff
@@ -1,18 +1,95 @@
...
+
+      // app details
+      const acronymObj = await acronyms.getAcronym(sc.acronym);
+
+      // if a corresponding acronym in db
+      if (acronymObj) {
+        sc.acronymDetails = acronymObj.dataValues;
+
+        // promotions (from lifecycle table for now)
+        const promotions = await lifecycleService.findLatestPromotions(sc.acronymDetails.acronymId);
```
loneil, line 29: Discussed refactoring this so we're not doing db calls per iteration in the loop. Can live with it for now but let's do a backlog ticket to track a performance enhancement. Might not ever be important but if this ever gets used outside of the admin context, or starts getting slow with rising numbers of clients we'll have to address.

app/src/components/clients.js
```diff
@@ -1,18 +1,95 @@
...
+
+        promotions.forEach(promotion => {
+          if (promotion.length) {
+            const data = promotion[0].dataValues;
+            sc.environments[data.env].created = data.createdAt;
+            sc.environments[data.env].updated = data.updatedAt;
+          }
+        });
+
+        sc.users = await acronyms.getUsers(sc.acronym);
```
loneil, line 39: Same as above, db calls in loop.

app/src/services/lifecycle.js
```diff
@@ -33,4 +33,39 @@ module.exports = {
...
+   * @returns {object} A lifecycle object if it exists
+   */
+  async getLatestEnvAction(acronymId, env) {
+    const lifecycle = await db.Lifecycle.findAll({
+      where: {
+        acronymId: acronymId,
+        env: env
+      },
+      order: [
+        ['updatedAt', 'DESC']
```
loneil, line 51: This will work I think (since in this case updatedAt will always be the same), but createdAt would be the relevant field here if I'm not mistaken?

app/frontend/src/components/admin/ServiceClients.vue
```diff
@@ -0,0 +1,234 @@
...
+          <td :colspan="headers.length - 1" :class="[responsiveCell]">
+            <div class="kc-expanded">
+              <!-- app details -->
+              <strong>Name: </strong>
+              <span>{{ item.name }}</span>
+            </div>
+
+            <div class="kc-expanded">
+              <!-- app details -->
+              <strong>Acronymn Created: </strong>
```
loneil, line 77: typo

app/src/services/lifecycle.js
TimCsaky: I guess so. especially if a password reset would match a record.
i can change that now to created.

app/frontend/src/components/admin/ServiceClients.vue
```diff
@@ -0,0 +1,234 @@
...
+  display: none !important;
+}
+tr.v-data-table__expanded__content
+  td.v-data-table__mobile-table-row:not(:nth-child(1)) {
+  padding: 0;
+}
+td.v-data-table__mobile-table-row div.kc-expanded {
+  padding: 0.2rem 1rem;
+}
+</style>
```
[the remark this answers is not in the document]
TimCsaky: I was thinking the same thing but read something about sass not handling >>> properly. I'll look into that again later.

app/src/components/clients.js
TimCsaky: thanks for pointing this out. i should have really caught that.

app/src/services/lifecycle.js
TimCsaky: yeah i assumed either was good. we're calling it 'environment last updated' in the ui.

app/frontend/src/components/admin/ServiceClients.vue
```diff
@@ -22,7 +22,7 @@
         class="kc-table"
         :headers="headers"
         :items="serviceClients"
-        :items-per-page="10"
+        :items-per-page="7"
```
loneil, line 25: This change doesn't seem to be working. I think it only takes 5, 10, 15...
10 is better IMO too

TimCsaky: ok. sounds good.

TimCsaky: i know you suggested making them alphabetical too. but i didnt get that done yet

jujaga: Let's leave the alphabetical stuff out for now.

PR #170, opened 2020-04-07
PW should reset for existing clients

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-954

Should reset the password when updating an existing service client.

Added a call to 'manage' in the KC service manager to do it, would be probably better to extract out to a specific 'reset' functionality, but for not since the UI is tied together it needs to be able to update the name/description at the same time.

## Types of changes

## Checklist

### DIFF
--- a/app/src/components/realmAdminSvc.js
+++ b/app/src/components/realmAdminSvc.js
@@ -231,6 +231,26 @@ class RealmAdminService {
     return response.data;
   }
 
+  async generateNewClientSecret(id) {
+    if (!id) {
+      log.error('RealmAdminService generateNewClientSecret id parameter is null.');
+      throw new Error('Cannot update client secret: id parameter cannot be null.');
+    }
+
+    const response = await this.axios.post(
+      `${this.realmAdminUrl}/clients/${id}/client-secret`,
+      {
+        headers: {
+          'Content-Type': 'application/json'
+        }
+      }
+    ).catch(e => {
+      log.error('RealmAdminService.generateNewClientSecret', JSON.stringify(e));
+      throw e;
+    });
+    return response.data;
+  }
+
   async getServiceAccountUser(id) {
     if (!id) {
       log.error('RealmAdminService getServiceAccountUser id parameter is null.');
--- a/app/src/components/realmAdminSvc.js
+++ b/app/src/components/realmAdminSvc.js
@@ -233,7 +233,7 @@ class RealmAdminService {
 
   async generateNewClientSecret(id) {
     if (!id) {
-      log.error('RealmAdminService generateNewClientSecret id parameter is null.');
+      log.error('RealmAdminService.generateNewClientSecret', 'RealmAdminService generateNewClientSecret id parameter is null.');
       throw new Error('Cannot update client secret: id parameter cannot be null.');
     }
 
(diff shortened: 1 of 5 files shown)


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-170

app/src/components/realmAdminSvc.js
```diff
@@ -231,6 +231,26 @@ class RealmAdminService {
     return response.data;
   }
 
+  async generateNewClientSecret(id) {
+    if (!id) {
+      log.error('RealmAdminService generateNewClientSecret id parameter is null.');
```
jujaga, line 236: log.error('RealmAdminService.generateNewClientSecret', 'id parameter is null.');

app/tests/unit/components/realmAdminSvc.spec.js
```diff
@@ -280,6 +280,39 @@ describe('RealmAdminService getClientSecret', () => {
...
 });
 
+describe('RealmAdminService generateNewClientSecret', () => {
+
+  it('should throw an error connection is not set', async () => {
+    const svc = new RealmAdminService(realmConfig);
+    svc.axios = undefined;
+    await expect(svc.generateNewClientSecret('1')).rejects.toThrow();
+  });
+  it('should throw an error connection is not set', async () => {
```
jujaga, line 290: These it description strings are the same between the two tests.

app/tests/unit/components/realmAdminSvc.spec.js
```diff
@@ -280,6 +280,39 @@ describe('RealmAdminService getClientSecret', () => {
...
+    await expect(svc.generateNewClientSecret(undefined)).rejects.toThrow();
+  });
+
+  it('should return a secret update response', async () => {
+    const svc = new RealmAdminService(realmConfig);
+    svc.axios = axios.create();
+    mockAxios.onPost(`${svc.realmAdminUrl}/clients/1/client-secret`).reply(200, 'truthy');
+
+    const result = await svc.generateNewClientSecret('1');
+    expect(result).toBeTruthy();
```
jujaga, line 312: Optional: When we're using mockAxios, it's worth adding in an extra expectation assertion to check that the mocked axios was actually invoked. This can be done something like...
```
expect(mockAxios.history.post).toHaveLength(1);
expect(mockAxios.history.post[0].data).toEqual('truthy');
```
Obviously would need to clear mockAxios state on a beforeEach to be consistent. (Ref utils.spec.js for example)

app/src/components/realmAdminSvc.js
loneil: Fixed

app/tests/unit/components/realmAdminSvc.spec.js
loneil: Fixed

app/tests/unit/components/realmAdminSvc.spec.js
loneil: k, will look at this whole test file later, following existing pattern for this quick bug

PR #173, opened 2020-04-17
Fix bug causing API Access tab actions to fail to load

loneil:
# Description
Introduced in last PR, I typo'd an existing parameter name somehow that broke to service client per acronym API call (yikes!).
Fixed with a rollback in prod so nothing broken at the moment.

## Types of changes

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-173

loneil:
@jujaga Thanks, this is what I get for forgetting a "toHaveBeenCalledWith" on the spy, would have caught this (adding in my future PR to all permission checks)

PR #176, opened 2020-05-06
Vue Skeleton backport updates

jujaga:
# Description

This PR applies the key vue skeleton backport changes to the codebase. Some key changes include:
- Keycloak libraries updated to 10.0.0 and remove outdated external dependencies
- Full npm package security updates for app and frontend
- Update base and bcgov components and respective tests
- Add Cypress e2e testing scaffold

[SHOWCASE-1100](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-1100)

## Types of changes

## Checklist


bcgov-nr-csst:
SonarQube Analysis: https://sonarqube-k8vopl-tools.pathfinder.gov.bc.ca/dashboard?id=nr-get-token-pr-176

app/frontend/src/components/base/BaseSecure.vue
```diff
@@ -32,15 +32,15 @@ export default {
...
       'keycloakReady'
-    ]),
-    authorized() {
-      return this.admin ? this.isAdmin : true;
-    }
+    ])
   },
   methods: {
+    authorized() {
+      return this.admin ? this.isAdmin : true;
```
TimCsaky, line 39: so does this mean only users in the 'Realm Administrator' group can see the secured ui?

jujaga: This authorized function is actually the same as the previous version where it existed as a computed attribute. This is only checking to see if a user has the GETOK_ADMIN role on the realm - this is different than the Realm Administrator role you're thinking about.

I moved this authorized function to methods because if/when we start having multiple roles we have to check on, it needs to be calculated during runtime and cannot be cached (ref silvipc authorized).

TimCsaky: ok. thanks. good to know. i'll take another look at that

app/frontend/src/main.js
```diff
@@ -92,7 +92,9 @@ function loadKeycloak(config) {
       realm: config.keycloak.realm,
       url: config.keycloak.serverUrl
     },
-    onReady: () => {},
+    onReady: () => {
+      initializeApp(true);
+    },
     onInitError: error => {
       console.error('Keycloak failed to initialize'); // eslint-disable-line no-console
```
parc-jason, line 99: this should probably raise an error or intializeApp(false).  we only do something on the happy path and nothing on error.

app/frontend/src/components/base/BaseSecure.vue
parc-jason, line 39: if authorized is only the in the case where admin is true, then shouldn't this be called admin()?  are we saying the the deault use of this component is for admins only? but other components could change authorized() to check if some other value(s) were true?

jujaga: I left this as authorized() because this function is actually intended to check for resource roles in the long run. At this time, Getok is tied to realm roles instead, so looking for the correct roles has to look at the realm level. The overall intent is that this authorization function would look more along the lines of what is in silvipc (refer to the BaseSecure implementation version) which is why the function is still named authorized. However in this case, the implementation retains the simpler isAdmin check to retain current functionality.

We're going to have to quickly decide on getok if we want to swap from realm level to resource level roles, or if we want to leave it as is before it gets developed too much further.

app/frontend/src/main.js
jujaga: I've tested the case where config is loaded, but it has incorrect Keycloak data in order to trigger the onInitError. As this is an asynchronous execution, we don't particularly have to re-execute initializeApp again because the loadConfig function already does this for us. When I added initializeApp to onInitError, it actually caused Vue to completely reload twice, which is unnecessary.

openshift/README.md
```diff
@@ -2,7 +2,7 @@
 
 This application is deployed on Openshift. This readme will outline how to setup and configure an Openshift project to get the application to a deployable state. There are also some historical notes on how to bootstrap from nothing to fully deployed on Openshift. This document assumes a working knowledge of Kubernetes/Openshift container orchestration concepts (i.e. buildconfigs, deployconfigs, imagestreams, secrets, configmaps, routes, etc)
 
-Our builds and deployments are orchestrated with Jenkins (see [tools](../tools/jenkins/README.md)). Refer to [Jenkinsfile](../Jenkinsfile) and [Jenkinsfile.cicd](../Jenkinsfile.cicd) to see how the Openshift templates are used for building and deploying in our CI/CD pipeline.
+Our builds and deployments are orchestrated with Jenkins. Refer to [Jenkinsfile](../Jenkinsfile) and [Jenkinsfile.cicd](../Jenkinsfile.cicd) to see how the Openshift templates are used for building and deploying in our CI/CD pipeline.
```
loneil, line 5: Does this one use the common jenkinsfile? Can't remember if we'd want to reference it

jujaga: Yes - current getok is using the commonPipeline.groovy file approach now. If we need that referenced in readme I can address it in subsequent pr.

PR #314, opened 2021-11-05
Add history page and route

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-778

Use the new Deployment History table to display a history tab on the acronym front end.

## Types of changes

## Checklist

## Further comments

image

### DIFF
--- a/app/frontend/src/components/History.vue
+++ b/app/frontend/src/components/History.vue
@@ -16,7 +16,7 @@
           An error occurred while loading history, please refresh and try again
         </v-alert>
         <v-simple-table class="getok-simple-table">
-          <template v-slot:default>
+          <template>
             <thead>
               <tr>
                 <th class="text-left">Date</th>
@@ -29,6 +29,9 @@
                 <td>{{ formatDate(item.date) }}</td>
                 <td>{{ item.idir }}</td>
                 <td>{{ item.env }}</td>
+              </tr>
+              <tr v-if="histories.length === 0">
+                <small>No Service Client deployments have occurred yet for application <strong>{{ acronym }}</strong>.</small>
               </tr>
             </tbody>
           </template>
--- a/app/frontend/src/components/History.vue
+++ b/app/frontend/src/components/History.vue
@@ -1,0 +1,90 @@
+<template>
+  <v-container>
+    <v-skeleton-loader
+      type="table"
+      :types="{
+        table: 'table-thead, table-tbody',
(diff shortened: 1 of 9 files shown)


bcgov-nr-csst:
PR Deployed at: https://getok-dev.apps.silver.devops.gov.bc.ca/pr-314

app/frontend/src/components/History.vue
```diff
@@ -0,0 +1,90 @@
...
+      :loading="!loaded"
+      transition="scale-transition"
+    >
+      <v-container>
+        <p class="text-right">Application <strong>{{ acronym }}</strong> created on: {{ formatDate(acronymCreated) }}</p>
+        <v-alert v-if="errorLoading" type="error" tile dense>
+          An error occurred while loading history, please refresh and try again
+        </v-alert>
+        <v-simple-table class="getok-simple-table">
+          <template v-slot:default>
```
jujaga, line 19: Can't remember, but if it's the default v-slot, you might be able to omit the `v-slot:default` part.
Also wondering if it would be worthwhile putting in an "empty" message when no records show on table instead?

app/src/services/deploymentHistory.js
```diff
@@ -60,5 +60,30 @@ module.exports = {
...
+  },
+
+  /**
+   * @function findHistory
+   * Returns an array of the deployment history for an acronym, joined with the user records
+   * @param {string} acronym The desired acronym
+   * @returns {object[]} An array of history object
+   */
+  async findHistory(acronym) {
+    return await db.Acronym.findOne({
```
jujaga, line 72: Do we need the async/await wrapper here? Looks like a singleton operation so could just return the promise of the findOne operation instead (can't remember the general pattern we did in getok).

jujaga (commented):
Looks pretty good. Some minor comments for consideration.

app/src/services/deploymentHistory.js
loneil: All the other services look like they await the query and return that result so probably would stick with that for this one too.

app/frontend/src/components/History.vue
loneil: Yeah slot not needed.
No `no-data` prop on the v-simple-tables we use here so `v-if`'d a message in

jujaga (approved):

TimCsaky (approved):

